Issue #111, opened 2016-07-03
Use ElasticSearch python API to do Add/Delete data


DaleMcGrew:
Please create examples using the python ElasticSearch API (version 1.x) to do the following functions, so we can call the ElasticSearch API (from the WeVoteServer API server) when our data changes:
- [x] Add (or update if it exists) one entry for one table (ex/ Add a candidate) — this can be by “id” key
- [x] Delete one entry for one table (ex/ Delete a candidate) — this can be by “id” key

josephevans:
For adding documents to ES, the code should be something like this:

es = Elasticsearch(["es-ip:9200"], timeout = 20, max_retries = 5, retry_on_timeout = True)
result = es.index(index = "candidates", doc_type = "candidate", id = <unique id>, body = {
}
`es = Elasticsearch(["es-ip:9200"], timeout = 20, max_retries = 5, retry_on_timeout = True)
result = es.index(index="candidates", doc_type="candidate", id=<unique id>, body={
    "candidate_name": value, "candidate_twitter_handle": value, "twitter_name": value,
    "google_civic_election_id": blah, "state_code": blah, "we_vote_id" : blah
}`

To delete documents from ES:

`es.delete(index="candidates", doc_type="candidate", id=<unique id>)`

DaleMcGrew:
Thank you @josephevans, we will add this to the WeVoteServer application code.

udii:
I experimented with using django signals to get notification for objects saves and deletes (see [here]<https://github.com/udii/WeVoteServer/blob/notify-elastic-search/signals/elasticsearch.py>,
still need to find a better way to initialize the module and to access the instance,
before I can hook this up to elastic search.

DaleMcGrew:
Ready for test. Thank you @udii!

DaleMcGrew:
It works great @udii. Thank you.

PR #123, opened 2016-09-04
Notify elastic search

udii:
Use ElasticSearch python API to do Add/Delete data #111
https://github.com/wevote/WeVoteServer/issues/111


DaleMcGrew:
👍

PR #125, opened 2016-09-05
Added positionListForVoter API endpoint so we can show the voter's positions before they sign in.

DaleMcGrew:
Added positionListForVoter API endpoint so we can show the voter's positions before they sign in. Also added ability to see positions if you sign in with Facebook. Renamed PositionEnteredManager to PositionManager.


DaleMcGrew:
👍

PR #138, opened 2016-09-22
Change twitter TweepError handling

eshyong:
There was an error in the way we were handling TweepErrors from the twitter library. Apparently the errors returned from the API no longer contain dicts but instead contain TweepErrors, so python was throwing an exception like:

```
TypeError: 'TweepError' object is not iterable
```

To solve this, we don't iterate over the object but append TweepError.reason to our returned status string.


DaleMcGrew:
👍

PR #150, opened 2016-10-06
Check in of friending code for broader testing. Adding code to transfer data from one voter account to another.

DaleMcGrew:


DaleMcGrew:
👍

PR #151, opened 2016-10-07
Check in of Facebook sign in re-vamp. Adding code to transfer data from one voter account to another.

DaleMcGrew:


DaleMcGrew:
👍

PR #154, opened 2016-10-11
Improved the way we handle friend invitation links. Now able to see process invitations on Requests page. Added a way to turn off SendGrid sending when testing.

DaleMcGrew:


DaleMcGrew:
👍

PR #155, opened 2016-10-14
Added to the mergeTwoVoters routine: now email addresses are moved to new account. Upgraded the Facebook sign in process to be less error prone.

DaleMcGrew:


DaleMcGrew:
👍

PR #160, opened 2016-10-17
Updated Office, Measure and Candidates to check for existing google_civic identifier before creating a new entry.

DaleMcGrew:


DaleMcGrew:
👍

PR #161, opened 2016-10-17
Fixed Office, Candidate and Measure models to deal with identically named offices that are in different districts.

DaleMcGrew:


DaleMcGrew:
👍

PR #163, opened 2016-10-18
Fixed issue with saving to unique fields. Better support for deleting emails and making a specific email the primary. Added state_code to many drop downs.

DaleMcGrew:


DaleMcGrew:
👍

PR #166, opened 2016-10-20
Origin/develop

josephevans:
If a politician result is received from ElasticSearch, skip over it instead of stopping iteration of the results.


DaleMcGrew:
👍

PR #167, opened 2016-10-20
Added remove_voter_cached_email_entries_from_email_address_object to prevent voter.email and voter.primary_we_vote_id collisions. Fixed Twitter sign in problem on second sign in. Fixed display bug caused by candidate party being None.

DaleMcGrew:


DaleMcGrew:
👍

PR #183, opened 2016-11-02
Fix to organizationSave infinite loop. Create a linked organization when you accept friend email invitation.

DaleMcGrew:


DaleMcGrew:
👍

PR #186, opened 2016-11-03
Removed people who shouldn't be suggested as friends. Updated import ballots code to support a "state_code" variable.

DaleMcGrew:


DaleMcGrew:
👍

PR #191, opened 2016-11-05
When we get lat/long for ballot_returned, we want to also save it to polling_location table so we don't have to look lat/long up for future elections. Limiting election summary page by state or 1000 records.

DaleMcGrew:


DaleMcGrew:
👍

PR #197, opened 2016-11-09
Added state_code to various admin pages so we can run reports on 2016 election. Including Support/Oppose figures.

DaleMcGrew:


DaleMcGrew:
👍

PR #202, opened 2016-11-23
Heal organization names that are meant to be temporary. Created bulk way to create new organizations. Removed dependence on voter.facebook_id. Added a routine to refresh position data for one organization's positions. Now repairing missing google_civic_election_id for positions for measures.

DaleMcGrew:


DaleMcGrew:
👍

PR #203, opened 2016-11-23
Repairing position names to match organization name. Now adding a voter guide for organizations if it is missing when you visit the organization's page.

DaleMcGrew:


DaleMcGrew:
👍

PR #216, opened 2017-02-01
Refactored the organization_follow controller code to use existing toggle_on_voter_following_organization function.

DaleMcGrew:


DaleMcGrew:
👍

PR #224, opened 2017-02-11
Adding voter guides to follow ballot_item-by-ballot_item. Added list of ballot items that an organization has an opinion about for display on main ballot page.

DaleMcGrew:


DaleMcGrew:
👍

PR #243, opened 2017-03-03
All references to star changed to bookmark (candidate campaign, conte…

mnneelam:
…st office and contest measure)

-  Added app bookmark, copy of Star app
-  Added url for bookmark in config file
-  Added find_and_update_bookmarks_view to copy rows from StarItem table to BookmarkItem table
-  Fixed default value from 0 to None for contest_office_id, contest_measure_id and candidate_campaign_id in bookmark models file


DaleMcGrew:
Thank you for doing all of this work @mnneelam! 👍

PR #248, opened 2017-03-07
Changes related to image width and height in caching resized image

jainanisha90:
Changed image width and image height to string in cache and resized cache image function.
Added delete we vote image function when error occurs during caching and resizing


DaleMcGrew:
👍

PR #249, opened 2017-03-07
Replace references to serializer with values() and later create JSON …

mnneelam:
…for all SyncOutView APIs

1) Changed SyncOutView APIs from class-based view to function based view
2) use of values() to create JSON instead of serialize
3) commented out export SyncOutView function from app url file
4) Updated apis_v1 url file with function based view
5) removed format JSON from API documentation for SyncOut APIs


DaleMcGrew:
Thank you @mnneelam! 👍

PR #253, opened 2017-03-14
voterUpdate API endpoint and some other APIs related to voter images updated with respect to cached all size image urls

jainanisha90:


DaleMcGrew:
👍

PR #256, opened 2017-03-16
Added "stripe" to requirements.txt. Comments about deprecating two functions.

DaleMcGrew:
Added "stripe" to requirements.txt. Comments about deprecating two functions.


DaleMcGrew:
👍

PR #257, opened 2017-03-16
Updated “organizationsFollowedRetrieve” to return organizations you are following because of Twitter sign in

jainanisha90:


DaleMcGrew:
👍

Issue #263, opened 2017-03-23
Resized profile images: Issue with image ratio


DaleMcGrew:
Some Twitter profile images are not square. Currently when we resize profile images, we assume they are square, so the resulting image looks squished. See https://twitter.com/jenancona for example.
- [ ] If the full sized profile image is not square, crop the right and left sides so the full sized image is square before resizing.
<img width="952" alt="screen shot 2017-03-22 at 10 27 38 pm" src="https://cloud.githubusercontent.com/assets/7756031/24233496/dae74a1a-0f4f-11e7-913b-065d1b34399a.png">

- [ ] Make sure to not crop original images (even if profile image), nor background or banner images
- [ ] Make sure to maintain the image ratio when you resize background or banner images (so they don't get resized into squares

DaleMcGrew:
Hi @jainanisha90 I just realized that your pull request is also cropping the background and banner images. I added more specification above to the original description.

DaleMcGrew:
To be verified.

DaleMcGrew:
Hi @jainanisha90, on this page: https://api.wevoteusa.org/org/1521/pos/
When I click "refresh Twitter details" the image does not get updated. Could we figure out a way to reset this image?

DaleMcGrew:
Fixed!

PR #265, opened 2017-03-27
Fixed issue #263 with image ratio during resizing profile images and Update facebookFriendsAction API to return facebook friends list with facebook suggested friends list

jainanisha90:


DaleMcGrew:
👍

Issue #267, opened 2017-03-28
No need to resize background or banner images to medium or tiny sizes


DaleMcGrew:
It is ok to resize background or banner images to the large photo height, but there is no need for medium or tiny sizes of any background or banner images.

DaleMcGrew:
Thank you @jainanisha90!

Issue #269, opened 2017-03-29
Make sure we grab Twitter photos in create_resized_images_for_voters function


DaleMcGrew:
Upgrade image/controllers.py create_resized_images_for_voters. If a voter is linked to a Twitter account in the "TwitterLinkToVoter" table, and we don't have Twitter photos for this voter yet, we should reach out to the Twitter API and cache the images in our Images table.

DaleMcGrew:
Great work @jainanisha90!

Issue #268, opened 2017-03-28
Retrieve photo based on Facebook Id, even if not in FacebookAuthResponse table


DaleMcGrew:
Upgrade image/controllers.py create_resized_images_for_voters. If a voter is linked to a Facebook account in the "FacebookLinkToVoter" table, but there isn't an entry in the FacebookAuthResponse, we should grab a photo directly from the default Facebook Photo URL and cache it in our Images table.

DaleMcGrew:
Hello Anisha, when I go to this voter on the live database, I see that he signed in with Facebook. When I use "create_resized_images_for_this_voter" from this page https://api.wevoteusa.org/image/104799/images_for_one_voter/, it does not create images for him From his Facebook profile photo.

https://api.wevoteusa.org/voter/104799/edit/
Facebook ID: 10210195462217373

DaleMcGrew:
Verified!

PR #271, opened 2017-03-29
Fixed issues #267 #268 #269

jainanisha90:
- #267 No need to resize background or banner images to medium or tiny sizes
- #268 Retrieve photo based on Facebook Id, even if not in FacebookAuthResponse table
- #269 Retrieve photo based on twitter ID even if not in twitter user table


DaleMcGrew:
👍

PR #275, opened 2017-04-04
Stripe api integration

sarah-clements:
Created managers to query or create stripe customer id's for processing charges, to create single payment charges and save in database, to query or create recurring donation plans and to create and save recurring subscriptions - fully functioning single payment and recurring donations.


DaleMcGrew:
👍

PR #277, opened 2017-04-07
Added support for CTCL data  import

mnneelam:
1) Added app for electoral district and party
2) Updated import_export_ctcl app with import for party and entire CTCL data
3) Added script to import_export_ctcl template for importing CTCL data in a given sequence
4) fixed a few warnings


DaleMcGrew:
Great work @mnneelam, this is some excellent code. 👍

PR #278, opened 2017-04-12
Stripe API - specific error handling

sarah-clements:
Created a dict, controller function and model function for generating a specific error message based on http status and error type; failures due to card errors will be generated based on key/value entries in the card_error_message dict.


DaleMcGrew:
Looks great! 👍

PR #291, opened 2017-05-01
Steve wevoteserver may1

SailingSteve:
In apis_v1/documentation_source/voter_retrieve_doc.py, fixed the order of the fields in the
api display, to match current return.  Added section at the bottom to show new voter_donation_history_list that shows each payment that was made.

In donate/controllers.py the donation_with_stripe_for_api now returns the donation with a call
to donation_manager.create_donation_history_entry — this is a new table that is specifically
for donations, and can retrieve a list of donations with a single query of a single table.
There is also a new method donation_history_for_a_voter that retrieves the list for a single voter.

In donate/models.py added some fields to the existing DonationSubscription table that allows the
data to be associated with the data in stripe, and to be displayed in a new subscriptions tab in the 
webapp.  Created new table DonationHistory to contain all the voters donation records, all the key 
stripe ids are stored, so we could debug using their data in some unusual case.  New function
create_donation_history_entry that stores a donation history record, and retrieve_donation_history_list to get the list from the db.

In voter/controllers.py added a call to donation_history_for_a_voter(voter.we_vote_id) in 
voter_retrieve_for_api(voter_device_id) to send the donation history to the WebApp

### DIFF
--- a/apis_v1/documentation_source/voter_retrieve_doc.py
+++ b/apis_v1/documentation_source/voter_retrieve_doc.py
@@ -49,17 +49,17 @@ def voter_retrieve_doc_template_values(url_root):
                    '  "voter_created": boolean,\n' \
                    '  "voter_found": boolean,\n' \
                    '  "we_vote_id": string,\n' \
-                   '  "first_name": string,\n' \
-                   '  "last_name": string,\n' \
-                   '  "email": string,\n' \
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
Great work! Congratulations on your first We Vote pull request. 👍

PR #292, opened 2017-05-02
donation_history_for_a_voter query now uses all() to retain keys in results set.

SailingSteve:
row references.
retrieve_donation_history_list now does a query all(), instead of
retrieving individual fields, so that the result set contains the keys.


DaleMcGrew:
👍

PR #293, opened 2017-05-02
Added interface_status_flags to voter record

bharathdn:
updated voter retrieve and voter update docs
Fixed bug on Election Syncout Api


DaleMcGrew:
Congratulations on your first We Vote pull request! 👍

PR #298, opened 2017-05-10
Various changes to get subscription info back in voterRetrieve

SailingSteve:
Updated voter_retrieve_doc.py
Pre-formatted some data in donation_history_for_a_voter, so that I don't
have to do it on the React side.
donate/models.py commented out an unused db table, replaced the week old
DonationHistory table with a similar DonationJournal table that handles
onetime and subscription records.


DaleMcGrew:
👍

PR #300, opened 2017-05-11
Bug fixes to Sync Data With Master Server 

bharathdn:
controllers.py:
- added checks to set status on exception, in ballot_items_import

ballot models.py:
- added function to search for ballots by address, when geocoder fails
  to return a location
- removed iexact check from we_vote_id and voter_id
- formatted lot of tab spaces for readability

polling_location models.py:
 - added 'we_vote_id' to PollingLocationManager's result, so that
   a new we_vote_is is not created in DB

wevote_social utils.py:
- added checks before storing image cache


DaleMcGrew:
Thank you for these changes Bharath! 👍

PR #302, opened 2017-05-12
Renamed shortened variables, columns, and enums from subs to subscriptions

SailingSteve:


donate/models.py
```diff
@@ -300,7 +301,7 @@ def retrieve_stripe_customer_id(self, voter_we_vote_id):
                 else:
                     success = False
                     status = "EXISTING_STRIPE_CUSTOMER_ID_NOT_FOUND"
-            except Exception as e:
+            except Exception:
```
DaleMcGrew, line 304: Hi Steve, I like leaving the " as e" because when debugging you get more information from that variable. We also use that variable for adding error messages to status variables in a variety of places.

DaleMcGrew:
👍

SailingSteve:
Makes sense, I put them back in temporarily when I debug in those 
spots!  There is a irritating lint message that asks you to remove 
them.  I'll put this one back in.

Thanks,
Steve

Issue #305, opened 2017-05-16
Create move_donation_info_to_another_voter function in donate/controller.py


DaleMcGrew:
Please create a function in donate/controller.py called "move_donation_info_to_another_voter". Please call this function near the bottom of "voter_merge_two_accounts_for_api" in voter/controllers.py. This function should result in donations and donation history information that are spread across two voters to be merged into a single voter:
- [ ] Update all voter_we_vote_id values in the donation tables: DonateLinkToVoter, DonationFromVoter, DonationLog, and DonationSubscription.

Let's discuss problem cases that might arise.

SailingSteve:
Sure.  I would like to get basic Webhook data flowing first -- at that point all the data in postgres would stay current, including subscription expiration status.

SailingSteve:
This has been implemented

DaleMcGrew:
Hi Steve, I took a closer look at "move_donations_between_donors" in donate/models.py and it only seems to be moving entries from the DonationJournal. We need to move/update data from DonateLinkToVoter, DonationFromVoter, DonationLog, and DonationSubscription.

DaleMcGrew:
Thank you @SailingSteve for your recent check in that finishes cancelling donations and refunding subscriptions. Will you have time to work on this issue next?

SailingSteve:
Still cleaning up existing code, but I will get to this next.

SailingSteve:
@DaleMcGrew 
* Over the last few weeks, I looked at the pre-existing schema, and thought about what I would need to do to make subscriptions and webhooks work, and in the process I significantly simplified the schema.  Hopefully not losing out on any potential capabilities.  I left all the pre-existing tables and code in place, while building up the new capabilities.  Then when the new table was working and meeting all needs, I pulled out the parts of the older code and tables that weren’t being used.
* The DonationJournal contains every bit of data we need, it has long rows, but you can find anything without a join or a multi-table lookup.  By being frugal with column widths, it takes up less disk space than the prior tables.
* The stuff I pulled out:
  * DonateLinkToVoter is still there, but is not read by the code -- I don’t think it ever was.
  * DonationFromVoter is gone.    
  * DonationLog is gone, superseded by DonationJournal that has a lot more columns.
[...]

PR #311, opened 2017-05-22
Added move_donations_between_donors re: server side issue #305

SailingSteve:
Preliminary work for Stripe Webhook -- still working with Stripe to resolve a response signature issue.


DaleMcGrew:
👍

PR #316, opened 2017-05-25
Changes for having Stripe journal records update in the brower as store changes

SailingSteve:
In views_voter.py in voter_update_view add the ability to send the
latest Stripe journal entries to the webapp if send_journal_list is sent
with not other parameters.
In donate/controllers.py and models.py, stripped out lots of unused code


DaleMcGrew:
👍  Great work Steve!

PR #325, opened 2017-06-01
Added ability to cancel a subscription

SailingSteve:
In views_donation setup for subscription cancellation from the WebApp,
then when the webhook comes through on the server, we update the db. The
refund part is partially implemented -- the button is still disabled.
In apis_v1/urls added donationCancelSubscription and donationRefund.
In environment_variables_template.json, added "STRIPE_REFUND_DAYS"
which should be set to the number 30, for 30 days allowed for refunds of
charges.
In donate/controllers no longer attempts to write a charge to the db
after the subscription is created, now it relies on the webhook to do
that.  The old way was working, and stopped last week, and it was the
wrong way to do it.  Now in compliance with the stripe subscription
lifecycle.


DaleMcGrew:
Thank you @SailingSteve, this is some great code! 👍

PR #334, opened 2017-06-07
Cleanup of donation code, prior to working on upgrade to the latest stripe api

SailingSteve:
In donate/controllers.py, handle the maximum of 25 active subscriptions
error case.  Harden the storage of status strings, to handle strings
that could exceed 255 chars.  Lint cleanup.
In donate/models.py, handle some error conditions better.


DaleMcGrew:
👍

PR #337, opened 2017-06-08
Admin issues when loading ballots etc.

SailingSteve:
In election/controllers.py, we were determining api requests success
based on http success, not returned data success.  Fixed that.
In voter_guide/models.py we were referencing a variable that had not
been initialized, and were crashing.


DaleMcGrew:
👍

PR #338, opened 2017-06-09
In google civic, if the voter address doesn't parse, log an error

SailingSteve:
instead of failing silently.  This is a first step, we should have some
UI for this condition.


DaleMcGrew:
👍

PR #339, opened 2017-06-12
Python 3.6 logging library requires args and kwargs when just sending a string.

SailingSteve:
In lib/python3.6/logging/__init__.py
    def error(self, msg: object, args: object, kwargs: object) -> object:

If you are only sending in a string, you need to send {} for both *args
and **kwargs, or else you get a

[2017-06-12 22:24:33,988] [ERROR] Steve-Podells-MacBook-Pro.local:donate.controllers: donation_process_charge, general: error() missing 2 required positional arguments: 'args' and 'kwargs'

and if the logger.error call is in an exception code block, it throws nested exceptions.

### DIFF
--- a/candidate/controllers.py
+++ b/candidate/controllers.py
@@ -29,7 +29,7 @@ def candidates_import_from_sample_file():
     :return:
     """
     # Load saved json from local file
-    logger.info("Loading CandidateCampaigns from local file")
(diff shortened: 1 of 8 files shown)


DaleMcGrew:
Thank you Steve for catching this! 👍

PR #341, opened 2017-06-13
Added support for processing of candidate ctcl data

mnneelam:


DaleMcGrew:
👍  Thank you for working through all of these import types!

PR #349, opened 2017-06-20
Started off with this issue:

SailingSteve:
Google civic returns the same google_civic_election_id for primaries
that occur in multiple states on the same day, so we need to persist
the state_code most places where we store google_civic_election_id.
Added (working) error handling when receiving un-successful API queries.
Error response in these error conditions provides text that describes
the problem, instead of silently failing.

There was so much inconsistently "repeated" code, that resulted in
silently ignored errors, so I added process_request_from_master().

Reworked sync_data_with_master_dashboard.html so that it was more
understandable by newbies, and so that errors bubbled up to the
web page (instead of failing silently).

If I knew at the beginning how big this was going to become, I would 
have broken it up into multiple PRs.

I carefully inspected (many times) the code that runs on the master server 
and loads the original master copy, but could not fully test it on my local.

This change requires a DB migration.

### DIFF
--- a/ballot/controllers.py
+++ b/ballot/controllers.py
@@ -20,7 +20,7 @@ import requests
 from voter.models import BALLOT_ADDRESS, VoterAddressManager, \
     VoterDeviceLinkManager
 import wevote_functions.admin
-from wevote_functions.functions import convert_to_int, positive_value_exists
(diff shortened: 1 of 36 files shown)


DaleMcGrew:
@SailingSteve, thank you for taking on this Herculean task of adding state_code! I have been through all of the code, and it looks great. I will also run some tests locally. 👍

PR #352, opened 2017-06-21
Fixed python logging

SailingSteve:
Logging controlled by LOG_FILE_LEVEL now works, so we can have other 
than ERROR level log lines (DEBUG, WARNING, INFO etc) if so configured with LOG_FILE_LEVEL.
Now that log lines other than ERROR work, it became clear that the non-ERROR lines don't need the two empty dictionaries passed with the call.


DaleMcGrew:
Looks good @SailingSteve, thank you for cleaning this up. 👍

PR #353, opened 2017-06-22
Replaced many print() calls with logger.debug or info. Plus avoid one bad exception.

SailingSteve:
In views_admin.py, we were not passing the request to
elections_import_from_master_server() which threw an exception since
logging now works.


DaleMcGrew:
👍

PR #355, opened 2017-06-24
Fixed import saved ballots to save the google_civic_election_id

SailingSteve:
Also added some prints() for developers to show that the 50 minute process is making progress.

### DIFF
--- a/import_export_google_civic/controllers.py
+++ b/import_export_google_civic/controllers.py
@@ -643,6 +643,8 @@ def store_one_ballot_from_google_civic_api(one_ballot_json, voter_id=0, polling_
         polling_location = results['polling_location']
         polling_location_latitude = polling_location.latitude
         polling_location_longitude = polling_location.longitude
+    else:
+        logger.info("No location found for " + str(results))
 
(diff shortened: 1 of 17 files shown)


SailingSteve:
In update_or_create_ballot_returned, made sure that we have a google_civic_election_id in the saved record -- this was the main fix.

import_export_google_civic/controllers.py
```diff
@@ -659,6 +661,7 @@ def store_one_ballot_from_google_civic_api(one_ballot_json, voter_id=0, polling_
                     voter_address_dict,
                     election_date_text, election_description_text,
                     google_civic_election_id, state_code, voter_id, '')
+                # TODO: 6/25/17, If we store a ballot without latidude and longitude, it will not be found by a geo search.  Is this intentional?  Does the data have value?
```
DaleMcGrew, line 664: We need to save ballot_returned entries even if we don' t have lat/long because we can exceed our Google API limits. We may also not want to "burn through" our available geocode API calls. That is why we don't always try to geocode when we don't have a latitude/longitude.

DaleMcGrew:
👍

PR #356, opened 2017-06-27
Fix exception in error logging when we get an address that fails to parse.

SailingSteve:
Also reducing this in importance to an INFO logging


DaleMcGrew:
👍

PR #357, opened 2017-06-27
Fix exception in exception when Stripe webhook signature verification fails.

SailingSteve:
Also added 3 log lines for testing the master logging/splunk, which
will be removed in a day or two.


DaleMcGrew:
👍

PR #359, opened 2017-06-27
logger.error() call fixes.

SailingSteve:
In Python 3.4 (what we are still using on the Master application servers) and in the latest Python 3.6.1, logger.error("some string") works fine.
In Python 3.6, that same logger.error would need to pass empty arrays
for args and kwargs... logger.error("some string", {}, {})
Upgraded my local to 3.6.1, and made code changes to remove all of those
empty arrays for logger.error() calls in the codebase.
Changed calls from deprecated logger.warn() to logger.warning()


DaleMcGrew:
👍

PR #363, opened 2017-06-28
Added support for CandidateSelection mapping table - map between Candidate and Contest Office,

mnneelam:
updated BatchRow handling for Contest Office with candidate selection data


DaleMcGrew:
👍

PR #365, opened 2017-06-29
Fix for the missing last4 on subscriptions.  

SailingSteve:
Some preliminary work for saving "Last Charged" in the db.
Requires a "python manage.py migrate"


DaleMcGrew:
👍

PR #367, opened 2017-06-30
add additional field to organizationFollow api and correct sort formula in issuesRetrieve

bharathdn:
add additional field to organizationFollow api and correct sort formula in issuesRetrieve


DaleMcGrew:
👍

PR #370, opened 2017-07-10
Adds a last_charged field to the client side donation subscription list.

SailingSteve:
Added a complicated way to get last_charged on the donation subscription list.

In donate/controllers, we now handle the Stripe invoice.created event.
In donate/models added a DonationInvoice table and new method to
cache the donation invoice in the db, and another to update the subscription
with the latest_charge date.

This change requires a db migration


DaleMcGrew:
👍

PR #372, opened 2017-07-11
Fixed a bug that caused crash when unlinking Twitter account.

DaleMcGrew:
Fixed a bug that caused crash when unlinking Twitter account.


DaleMcGrew:
👍

PR #374, opened 2017-07-12
ctcl changes - Added support for retrieving state_code info from BatchRowActionConte…

mnneelam:
…stOffice by doing a lookup using CandidateSelection mapping table data, updated import into CandidateCampaign from BatchRowActionCandidate with state_code, added a function to retrieve BatchRow column name based on BatchHeaderMap column value

### DIFF
--- a/import_export_batches/controllers.py
+++ b/import_export_batches/controllers.py
@@ -21,6 +21,8 @@ from office.models import ContestOffice, ContestOfficeManager, ElectedOffice, El
 from politician.models import Politician, PoliticianManager
 from electoral_district.controllers import retrieve_electoral_district
 from exception.models import handle_exception
+from django.db.models import Q
+
 
 logger = wevote_functions.admin.get_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


import_export_batches/controllers.py
```diff
@@ -993,7 +1064,7 @@ def create_batch_row_action_candidate(batch_description, batch_header_map, one_b
         # for this header_id (Duplicate entries in the same data set
         existing_batch_row_action_candidate_query = BatchRowActionCandidate.objects.all()
         existing_batch_row_action_candidate_query = existing_batch_row_action_candidate_query.filter(
-            batch_header_id=batch_description.batch_header_id, candidate_name=candidate_name,
+            batch_header_id=batch_description.batch_header_id, candidate_name=candidate_name, state_code=state_code,
```
DaleMcGrew, line 1067: Thank you @mnneelam, I am going to approve this pull request, but wanted to request that every time you filter for state_code, please include "__iexact" like this:
state_code__iexact=state_code
You might also want to use "__iexact" with the candidate_name.

DaleMcGrew:
👍

mnneelam:
Thank you @DaleMcGrew  for pointing out the __iexact ommission. I've modified the code to include __iexact  while filtering

PR #376, opened 2017-07-13
Bug Fixes to filter issues in issuesRetrieve and success flag when result list is empty

bharathdn:
Bug Fixes to filter issues in issuesRetrieve and success flag when result list is empty


DaleMcGrew:
👍

Issue #377, opened 2017-07-13
Twitter photos not retrieved on first sign in


DaleMcGrew:
When a person first signs in with Twitter, their photos are not retrieved. They are retrieved on the second sign in. We also want to make sure that the latest Twitter photos are being retrieved with each sign in if they have changed.

This drawing describes our Twitter sign in process: https://docs.google.com/drawings/d/1jzDld01ZvD8gguMkYK4GRwXW-smgb7kRXDIrIDTf68Y/edit

I think a good place to start is by searching for "twitterSignInRequestVoterInfo" in WeVoteServer. Our sign in system is complicated, so please don't hesitate to ask for help.

SailingSteve:
Clear from the db:
* twitter_twitteruser
* voter_voter (for my user)
* twitter_twitterlinktovoter
* twitter_twittercursortable
* voter_voterdevicelink (truncate)
* import_export_twitter_twitterauthresponse

Twitter photos not retrieved on first sign in #377

Then on the first login you get a database entry with the we_vote_hosted_profile_image_url_tiny with something like https://wevote-images.s3.amazonaws.com/wvbuvoter106/twitter_profile_image-20170716_1_24x24.jpeg in twitter_twitteruser, but not in voter_voter

voter_voter:
109,,,'wvbuvoter106','wvbuorg3',,,false,,,,'2017-07-15 20:23:03.404647-07','2017-07-15 20:23:53.819099-07',true,false,false,,,,,'856525338557399042',,'StevePodell',,,,,,,,,false,0,0

twitter_twitteruser
7,'856525338557399042','StevePodell','Steve Podell',,'https://wevote-images.s3.amazonaws.com/wvbuvoter106/twitter_profile_image_master-20170716_1_240x240.jpeg',,0,,,'https://wevote-images.s3.amazonaws.com/wvbuvoter106/twitter_profile_image-20170716_1_200x200.jpeg','https://wevote-images.s3.amazonaws.com/wvbuvoter106/twitter_profile_image-20170716_1_48x48.jpeg','https://wevote-images.s3.amazonaws.com/wvbuvoter106/twitter_profile_image-20170716_1_24x24.jpeg',

DaleMcGrew:
Thank you Steve for doing this research. So maybe we need a controller function to transfer image data stored in TwitterUser over to Voter after the first sign in is complete? The Amazon image is from our AWS image caching code.

Also I wonder why the Twitter banner isn't being stored in TwitterUser?

[...]

PR #383, opened 2017-07-17
Initial checkin for #377, now we save the photos on the first sign in.

SailingSteve:
(Still needs a webapp size change to take advantage of the photos)


DaleMcGrew:
@SailingSteve good catches! 👍

PR #384, opened 2017-07-19
Twitter photos and banner are available immediately after first sign in for issue #377

SailingSteve:
In import_export_twitter/controller, store the photos and banner in the
organization that is for a single user.
In import_export_twitter/models store the banner at auth time.
In organization/models added update_organization_single_voter_data()


SailingSteve:
This is now working.  If you change your profile picture, it updates.  If you change your banner, twitter keeps sending the old one... maybe it takes longer to update for the api?  I could pursue this issue with twitter if it is important.

DaleMcGrew:
👍

PR #385, opened 2017-07-19
CTCL changes for politician - state code support

mnneelam:
Added support for state_code in BatchRowActionPolitician by doing a lookup using PersonId in BatchRowActionCandidate table, added support for new XML nodes - State, Election and Source.


DaleMcGrew:
👍

Issue #378, opened 2017-07-13
Facebook banner not retrieved on sign in


DaleMcGrew:
We are currently not retrieving a person's Facebook banner on sign in. After retrieving the banner, I believe we have code the crops the Twitter banner that can be used? Or maybe we can retrieve the cropped banner from Facebook instead of the source image?

SailingSteve:
Also did a web side commit, that is part of the 378 fix.
Still unresolved:
- [x] Person image on /voterguide/ is constrained to a box, but should only be constrained by height, or investigate if we can get a big picture that is square.
- [x]  [2017-07-31 15:42:10,569] [ERROR] Steves-MacBook-Pro-2017.local:import_export_facebook.models: invalid literal for int() with base 10: 'false'
[2017-07-31 15:42:10,569] [ERROR] Steves-MacBook-Pro-2017.local:import_export_facebook.models: FACEBOOK_AUTH_RESPONSE_NOT_UPDATED_OR_CREATED, function: update_or_create_facebook_auth_response, file: /Users/stevepodell/PythonProjects/WeVoteServer/import_export_facebook/models.py, line: 239
- [x] [2017-07-31 15:41:32,526] [ERROR] Steves-MacBook-Pro-2017.local:import_export_facebook.models: An access token is required to request this resource.
[2017-07-31 15:41:32,526] [ERROR] Steves-MacBook-Pro-2017.local:import_export_facebook.models:  FACEBOOK_FRIENDS_LIST_NOT_FOUND, function: retrieve_facebook_friends_from_facebook, file: /Users/stevepodell/PythonProjects/WeVoteServer/import_export_facebook/models.py, line: 695
- [x] Figure out how to fork the logger.error on unhandled exceptions to go to the "disk" in addition to the http output stream.

SailingSteve:
Error logging when "SERVER_IN_DEBUG_MODE":  false resolved.
Error logging that previously was only going to HTTP 500 stream is resolved.
Twitter and Facebook banners now display on the VoterGuide page in the WebApp.

PR #395, opened 2017-07-31
We now store and display the facebook "cover" for issue #378

SailingSteve:
This change adds columns to the models and REQUIRES a db migration.
The python code base had, and almost certainly still has, unhandled
exceptions that are being redirected to the http stream, instead of
going to the the "disk" where they can be discovered by Splunk.  I made
some improvements in this area, and the errors now are also printed to
the Python console, but we need to figure out how to log to the "disk"
in parallel with logging to the http stream.
In views_organization, pass the voter_device_id into organization_retrieve_for_api
so that we can determine logged in state.
In views_voter pass through facebook_background_image_url_https to the
sign in code.
In import_export_facebook/controllers cache the facebook_background_image_url_https
and then store it with facebook_manager.create_or_update_facebook_user
In import_export_facebook add facebook_background_image_url_https to the FacebookAuthResponse
and facebook_email columns to the FacebookUser.  ADDED EXCEPTION HANDLING
for every update_or_create() .. this caught real undiagnosed uncaught exceptions.
In organization/controllers added a facebook data heal section to
organization_retrieve_for_api that initializes the Facebook data in the
organization.  organization_retrieve_for_api now sends either Facebook
or Twitter images to the client based on logged in state. In organization_save_for_api
we now also try harder to get correct Facebook data into the organization
object on save.
In Organization/models added column for facebook_background_image_url_https
In wevote_functions/admin, added a handler for uncaught exceptions
In wevote_functions/functions restored a print(message_text) line that
got deleted -- these lines have great value when waiting for multi-hour
sync with master server runs.
In wevote_social/middleware, added a fallback print to console "logger" for
unhandled exceptions that currently are not otherwise detectable except by line
by line debugging of function calls.

### DIFF
--- a/apis_v1/views/views_organization.py
+++ b/apis_v1/views/views_organization.py
@@ -61,8 +61,11 @@ def organization_retrieve_view(request):
     """
     organization_id = request.GET.get('organization_id', 0)
     organization_we_vote_id = request.GET.get('organization_we_vote_id', '')
+    voter_device_id = request.GET.get('voter_device_id', "")
     return organization_retrieve_for_api(
-        organization_id=organization_id, organization_we_vote_id=organization_we_vote_id)
+        organization_id=organization_id,
+        organization_we_vote_id=organization_we_vote_id,
+        voter_device_id=voter_device_id)
 
 
 def organization_save_view(request):  # organizationSave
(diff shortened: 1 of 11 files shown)


DaleMcGrew:
👍

PR #396, opened 2017-08-02
Refactor and bug fix for the facebook "cover" for issue #378

SailingSteve:
In update_or_create_facebook_auth_response() do not sent in a value for
facebook_user_id as a search criteria, it is already in the defaults.
In organization_retrieve_for_api() move the facebook heal code to voter
controllers voter_retrieve_for_api().


DaleMcGrew:
👍

PR #399, opened 2017-08-03
Bugfixes for the facebook "cover" for issue #378

SailingSteve:
In import_export_facebook/controllers.py in
voter_facebook_save_to_current_account_for_api, pass a dictionary to
retrieve_facebook_link_to_voter, which was previously failing with a
parameter mismatch.
In import_export_facebook/models, log at the debug level if no facebook
friends came back from the facebook api.  Change to come up with a safe
value for twitter_followers_count on Organization.create().  In
Organization.create() default to 0 for twitter_followers_count instead
of None, which would fail every time due to model column contraints.
Previously this was failing silently for logging to disk.
In voter/controllers fixed call to update_or_create_organization which
could on some conditions rely on an uninitialized organization_manager.
In Base.py STATIC_ROOT needed different settings when SERVER_IN_DEBUG_MODE=false
to build the correct path to serve apis_v1.css for the admin console.

### DIFF
--- a/config/base.py
+++ b/config/base.py
@@ -189,9 +189,13 @@ USE_TZ = True
 AUTH_USER_MODEL = 'voter.Voter'
 
 # Static files (CSS, JavaScript, Images)
-# https://docs.djangoproject.com/en/1.8/howto/static-files/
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

PR #401, opened 2017-08-04
Changes for the facebook "cover" for issue #378

SailingSteve:
This change set requires a database migration.
In apis_v1/views_organization we now receive a boolean from the client:
display_facebook_banner.
In import_export_twitter/controllers we now cache in aws, the twitter
banner and save link to the cached image in twitteruser.
In import_export_twitter/models the twitter auth response now contains
a value for the raw twitter banner.  In save_twitter_auth_values save
that raw banner, previously we were trying to save the value in a column
that did not exist.
In organization/controllers we now receive and act on that new boolean
from the client "display_facebook_banner", which allows us to avoid some
database calls that Dale did not like.


DaleMcGrew:
👍

PR #402, opened 2017-08-04
Error logging to "disk" for errors that previously were only sent to client issue #378

SailingSteve:
Added a new ADMINS setting/env variable that sends error level log events to a list of administrative emails. Leave it blank to disable. ADMINS is read by django/core/mail/__init__.py In Middleware.py, we now log error level lines to "disk" that previously were only getting sent to the client in the HTTP 500 stream.

### DIFF
--- a/config/base.py
+++ b/config/base.py
@@ -308,6 +308,10 @@ SOCIAL_AUTH_PIPELINE = (
 
 EMAIL_BACKEND = get_environment_variable("EMAIL_BACKEND")
 SENDGRID_API_KEY = get_environment_variable("SENDGRID_API_KEY")
+ADMINS = get_environment_variable("ADMINS")
+# Expecting a space delimited string of emails like "jane@wevote.org" or "jane@wevote.org bill@wevote.org"
+if ADMINS:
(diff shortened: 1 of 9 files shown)


config/base.py
```diff
@@ -308,6 +308,10 @@ def get_environment_variable(var_name, json_environment_vars=json_environment_va
 
 EMAIL_BACKEND = get_environment_variable("EMAIL_BACKEND")
 SENDGRID_API_KEY = get_environment_variable("SENDGRID_API_KEY")
+ADMINS = get_environment_variable("ADMINS")
+# Expecting a space delimited string of emails like "jane@wevote.org" or "jane@wevote.org bill@wevote.org"
+if ADMINS:
+    ADMINS = [[email.split('@')[0], email] for email in ADMINS.split()]
```
DaleMcGrew, line 314: Hi @SailingSteve, I learned the hard way on a large production system not to use email alerts for specific individual errors. Perhaps we could store in the database a count of errors seen in the system per day, and then email ADMINS daily the error count with a reminder to check Splunk? I look forward to discussing with you.

DaleMcGrew:
👍

PR #403, opened 2017-08-05
Updated new environment_variable to be named "ADMIN_EMAIL_ADDRESSES" instead of "ADMINS". Fixed API crash related to trying to run strftime on a "None" value. Opened up editing election data "election_name" and "election_day_text" to admins.

DaleMcGrew:
Updated new environment_variable to be named "ADMIN_EMAIL_ADDRESSES" instead of "ADMINS". Fixed API crash related to trying to run strftime on a "None" value. Opened up editing election data "election_name" and "election_day_text" to admins.


SailingSteve (commented):
Hi Dale,
   I just needed to add that ADMINS capability temporarily to debug the crashing error logging if sendgrid was not configured when in "SERVER_IN_DEBUG_MODE":  false, mode.  If ADMIN_EMAIL_ADDRESSES is "" in production it won't do anything, and also no problem if you want to take it out.  In lib/python3.6/lib/site-packages/django/core/mail/__INIT__.py line 91, it is looking for the existence of the ADMINS settings/global.  So as long as line 315 starts with ADMINS instead of ADMIN_EMAIL_ADDRESSES_ARRAY it should work, but as of now it is broken.

Thanks,
Steve

PR #410, opened 2017-08-11
Added Voter search function 

kotaaaa:
made a search box for voter.


DaleMcGrew:
Thank you for your first We Vote pull request @kotaaaa! 👍

Issue #427, opened 2017-08-24
How do we use a read-only database in Django?


DaleMcGrew:
We have a replicating AWS database in our server network. I would like to start using that end point for API calls that don't require any write activity. For this issue to be satisfied, please:

- [x] Propose changes to WeVoteServer/config/environment_variables-template.json for the read-only database
- [x] Update election/views_admin.py elections_sync_out_view to use the read-only database

SailingSteve:
What calls electionsSyncOut?  I couldn't find a reference in python or React.   If servers are using this to synchronize datasets, Postgres replication would do it much better.  (Probably orders of magnitude faster).

SailingSteve:
We want two sets of django dbs, and two sets of config for them in environment_variables.  Need to figure out how django discovers it's db, most likely on two different instances.

SailingSteve:
/Applications/Postgres.app/Contents/Versions/9.6/bin/pg_dump -p5433 WeVoteServerDB > backup.sql
Create WeVoteServerDbReadOnly with pgAdmin4
/Applications/Postgres.app/Contents/Versions/9.6/bin/psql -p5433 WeVoteServerDbReadOnly < backup.sql

SailingSteve:
Info on how to test this on two macs on your local network:

https://docs.google.com/document/d/1MJML7aT_QjSGtf_R1P9OpeILexRl5Ma5SZDiLuR5A68/edit?usp=sharing

SailingSteve:
I think this is complete.

DaleMcGrew:
Thank you @SailingSteve I have marked this as VerifyFix until we can test it on the production server.

DaleMcGrew:
This is working, thank you @SailingSteve!

PR #435, opened 2017-08-27
Server side #427, Update server to optionally use a read-only database.

SailingSteve:
For queries to election/views_admin.py elections_sync_out_view, use the replicated (identical) WeVoteServerDB on a slave designated as a readonly db.

In environment_variables-template.json add the block of settings for the
read only slave replicated postgres server.
In config/local.py add the DATABASES section for the read only server.
In controllers.py change the "Election.objects.all()" to
"Election.objects.using('readonly').all()" so that it uses the readonly
db server.

Each engineer on the project will need to make the changes to their
local environment_variables.json -- Add the block of 7 lines with the
_READONLY vars.  It is easiest to just copy all the settings from the
non readonly settings to the readonly.  In that way if you try to make
a request to your local on port 8000, you will get the read/write
database through the read/only settings.

This has been tested with two postgres servers running on two macs.

### DIFF
--- a/config/environment_variables-template.json
+++ b/config/environment_variables-template.json
@@ -4,13 +4,21 @@
   "WEB_APP_ROOT_URL":               "https://WeVote.US",
   "SERVER_IN_DEBUG_MODE":           true,
 
-  "_comment":                       "local.py settings, including database",
+  "_comment":                       "read/write database settings for local.py",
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
This is exciting @SailingSteve! Thank you for getting this working. I will also test locally, and then we will test it on the production servers. @josephevans @mshmsh5000 👍

PR #440, opened 2017-08-29
Fixed a voter guide bug

jainanisha90:
Retrieving all voter guides filtered by organization_we_vote_id method returning All voter guides without filter(Two functions have same name  in voter_guides/models.py, Python does not support function overloading)


DaleMcGrew:
👍

PR #443, opened 2017-08-30
add ability to search for positions and voter guides page

bharathdn:
add ability to search for positions and voter guides page


DaleMcGrew:
Looks great! Thank you @bharathdn. 👍

Issue #431, opened 2017-08-24
Receive Facebook photo offsets on server, and scale background image appropriately


DaleMcGrew:
We will need to update our image caching code to crop the banner (or reposition the banner within WebApp?), taking into consideration the offset_x and offset_y that Facebook returns to us.

SailingSteve:
This was called "Capture and pass Facebook photo offsets to WebApp"  am renaming it to "Receive Facebook photo offsets on server and scale background image appropriately"

This turned out to be a 100% server side issue, due to the messiness of doing cropping in React.

SailingSteve:
Done

DaleMcGrew:
Hi @SailingSteve, I just got a chance to do a code review. When I test it locally, I am not seeing my Facebook banner update, even after changing my banner on Facebook. Would you like to look at it together?

SailingSteve:
Works again for me! (LOL)

DaleMcGrew:
Fixed! Great work @SailingSteve.

PR #457, opened 2017-09-05
Server side #431, Persist and use facebook background offsets.

SailingSteve:
This is now an entirely server side implementation.
This change requires database migrations, and has two new environment
variables in the template:  "SOCIAL_BACKGROUND_IMAGE_HEIGHT" snf
"SOCIAL_BACKGROUND_IMAGE_WIDTH".
Receive Facebook background offsets from the client side call to
voter_facebook_sign_in_save_view() -- we were already sending the data,
but ignoring it serverside.
A big piece of the image caching chain now passes the two facebook
offsets all the way down to resize_we_vote_master_image() which uses
them to resize the facebook background image to the same size and aspect
ratio as the twitter background.
In image/controllers removed a bunch of useless code to make multiple
sizes for the facebook backgrounds.
In import_export_facebook/controllers, put the newly cached and sized
images into the organization object.  In display of the organization
banner and profile pictures, we unconditionally display the twitter
versions if they exist in the organization object.


DaleMcGrew:
Thank you @SailingSteve! 👍

Issue #452, opened 2017-09-01
Organization view: Highlight org type


mshmsh5000:
It should be clear in every view of an org (list, edit, position list, etc.) whether the org is a C3, C4, or other kind. These are already defined in the [model](https://github.com/wevote/WeVoteServer/blob/develop/organization/models.py#L18); they just need to be displayed prominently.

Especially important to reduce chance that a user will mistakenly attach a candidate position to a C3 instead of its related C4.

Find everywhere an organization is created, and assign the correct organization_type.
- [ ] If creating an organization entry due to action by voter, set to INDIVIDUAL

DaleMcGrew:
To be addressed in the future.

PR #458, opened 2017-09-06
Organization view: Highlight org type

bharathdn:
Organization view: provide ability to filter by organization type, update organization type

#452


DaleMcGrew:
Thank you @bharathdn, this will be very helpful. 👍

PR #459, opened 2017-09-06
#263 Added Delete Cached image option for Candidate, Organization and Voter

jainanisha90:
- Restore original twitter/facebook image urls in respected tables
- Delete cached images from aws
- Delete cached image entries from WeVoteImage Table


DaleMcGrew:
👍

PR #462, opened 2017-09-06
Server side #431, Persist and use facebook background offsets.

SailingSteve:
Fix to validating integer values for facebook offsets (plus an
accompanying client side fix should stop sending invalid values).


DaleMcGrew:
👍

PR #464, opened 2017-09-08
Adds instructions for prebuilt Vagrant box

mshmsh5000:
## What does this PR do?

Adds a README specifically for using the prebuilt Vagrant box, which is the equivalent of running the Ansible build from scratch.

Special thanks to @edward-ly for his patience as we got these instructions up to snuff.

## How can I test it?

Try following the instructions! You should end up with a WeVoteServer instance running at https://localhost:8889/


DaleMcGrew:
👍

PR #471, opened 2017-09-12
Update batch import key names for ballot items, various bug fixes

edward-ly:
Updated the ballot item batch import key names for consistency. Specified default values for state code and polling location search. Fixed some URL parameters to pass in variables and values correctly. Re-added links to edit polling location.


DaleMcGrew:
👍

PR #472, opened 2017-09-12
Unhandled exception on server side was causing grief in native iOS

SailingSteve:
Previously was returning a HTTP 500


DaleMcGrew:
👍

PR #494, opened 2017-09-22
#452 Fixed

jainanisha90:
- #452 
- Store twitter followers count and other twitter details when create twitter_user so that twitter followers count can be displayed on admin organization listing


DaleMcGrew:
👍

PR #499, opened 2017-09-25
oAuth support for react-native app

SailingSteve:
In urls.py, views_twitter, and import_export_twitter/controllers.py add
a new API for twitter_native_sign_in_save_for_api for react-native to
bypass the Webapp oAuth flow that is done server side with tweepy.
In native, steps 1 & 2are replaced with a react-native-oauth
authenticate() call.
In voter/models.py fixed an unhandled exception for IntegrityErrors


DaleMcGrew:
👍

PR #505, opened 2017-10-04
#1120 Identify request is coming from web-bots or human and from which device(like tablet, mobile or desktop)

jainanisha90:
- Used robot-detection for web-bots detection
- Used django-user-agents for device detection


DaleMcGrew:
👍

Issue #481, opened 2017-09-16
Import Ballots: "Polling Locations Picker"


DaleMcGrew:
Add a "Polling Locations Picker" to "Import Ballot at One Polling Location":
http://localhost:8001/import_export_batches/batch_list/?kind_of_batch=IMPORT_BALLOT_ITEM&google_civic_election_id=4389
 
![screen shot 2017-09-15 at 6 02 13 pm](https://user-images.githubusercontent.com/7756031/30507901-107ab722-9a40-11e7-822a-8d6ceb4f9c43.png)

You can copy it from:
http://localhost:8001/b/0/list_edit/?google_civic_election_id=4389
View code to copy: WeVoteServer/ballot/views_admin.py - search for "ballot_item_list_edit_view" and "ballot_item_list_edit_process_view"
Template code: WeVoteServer/templates/ballot/ballot_item_list_edit.html

PR #506, opened 2017-10-05
Implement Polling Locations Picker to Import Ballots

edward-ly:
Implemented the ability to select a polling location with a dropdown menu (and search by city or ZIP) rather than by location ID when importing ballots (#481).


DaleMcGrew:
👍 Thank you Edward!

Issue #482, opened 2017-09-16
Import Ballots: Analyze Data from csv


DaleMcGrew:
1. Go here: http://localhost:8001/import_export_batches/batch_list/?kind_of_batch=IMPORT_BALLOT_ITEM
2. Upload a Ballot file

On the left side of the "Import Ballot at One Location" screen, there is an "Analyze" link that takes you to "batch_action_list_analyze_process_view" in WeVoteServer/import_export_batches/views_admin.py

That view calls "create_batch_row_actions" in WeVoteServer/import_export_batches/controllers.py which needs code to be added for kind_of_batch=IMPORT_BALLOT_ITEM
I started a function called "create_batch_row_action_ballot_item" that you can finish.

You will also need to add IMPORT_BALLOT_ITEM-specific code to WeVoteServer/templates/import_export_batches/batch_action_list.html

DaleMcGrew:
This is done.

PR #523, opened 2017-10-11
WIP Issue #482

edward-ly:
Made progress into #482, to be finished at a later date.

Added a table when a IMPORT_BALLOT_ITEM batch row is analyzed (headers might not be correct). Added the IMPORT_BALLOT_ITEM case to kind_of_batch in create_batch_row_actions. Made partial progress into the create_batch_row_action_ballot_item function.


DaleMcGrew:
👍

PR #537, opened 2017-10-18
Updated algorithms for search and analysis of candidate's Twitter accounts

edward-ly:
Updated the searching of candidate's Twitter accounts by using additional search terms that factor in the candidate's initials and nickname (if any).
Updated the WeVote functions and CandidateCampaign model to extract all possible parts of a person's full name.
Updated the `nameparser` package to the latest version in pip.
Modified the scoring of each Twitter account to prioritize active accounts, correct names in the Twitter name and handle, and office names in the description. Scores can also increase or decrease depending on different keywords.
Modified the table headers of the Edit Candidate page to be consistent with those found in the search results.


DaleMcGrew:
👍

Issue #538, opened 2017-10-18
Possible Twitter Handles: Bulk Retrieve


DaleMcGrew:
Step 1:
Display the top result from "retrieve possible Twitter handles" (that was previously run) on http://localhost:8001/c/ (candidate/views_admin.py - candidate_list_view) in the "Twitter Handle" column. Show the top photo, Twitter handle, score and Twitter description. Make a click on the photo, Twitter handle, or score take you to the candidate_edit page.

Step 2:
On the candidate_list_view page, add a step between 3 & 4 called "Retrieve Possible Twitter Handles" that calls a view like this one: http://localhost:8001/twitter2/retrieve_possible_twitter_handles/wv01cand6714/
retrieve_possible_twitter_handles_view
...but that loops through 10 candidates at a time, and then returns to the candidate_list page.
Note: It should only retrieve possible Twitter handles if 1) there aren't any current entries, 2) A twitter handle isn't already saved.

DaleMcGrew:
Great work Edward!!

PR #544, opened 2017-10-25
Implemented bulk retrieval of possible Twitter handles

edward-ly:
Retrieved the top Twitter search result for each candidate without a linked Twitter account to display on the candidate list page. Implemented the ability to retrieve possible Twitter handles for multiple candidates at once. Refactored steps on candidate list page as unordered/ordered lists for ease of numbering.

To fix later: the bulk retrieval currently does not account for candidates with zero search results and so will no longer work if there are enough candidates with zero search results.

#538


DaleMcGrew:
👍  I love it!

PR #553, opened 2017-10-26
Implemented utc_offset check in Twitter search results analysis

edward-ly:
Implemented the ability to check if the utc_offset of a Twitter account is within two hours of the time zone of the current state. Added a helper function `convert_state_code_to_utc_offset` in wevote_functions.functions to handle the mapping from state_code to utc_offset. Re-added the ability to check if state_code is in the Twitter description since it is not already inside contest_office_name.


DaleMcGrew:
👍

Issue #549, opened 2017-10-26
Search Google Photos: Add "Choose and Update" button


DaleMcGrew:
Add to the results page a "Choose and Update" button under the photo on Edit Candidate page:
http://localhost:8001/c/wv01cand5748/edit/
<img width="646" alt="screen shot 2017-10-25 at 6 53 53 pm" src="https://user-images.githubusercontent.com/7756031/32031500-62a151c4-b9b6-11e7-9868-2e9281882a88.png">

When this button is clicked, we should cache the photo to We Vote's AWS photo store, and then add the photo to the candidate record.

DaleMcGrew:
Hi @jainanisha90, I tried the "Choose and Update" button on "Anthony Beckford" (New York), as well as a couple of other candidates, on the Google Search result from Twitter. It showed the green check mark, but did not save the Twitter handle. It did not seem to save the photo either. Save Twitter Link didn't work either.

DaleMcGrew:
I clicked "Choose and Update" for the Facebook search results for "Alan H. Reed":
https://api.wevoteusa.org/c/27715/edit/?show_all_google_search_users=1

...but it didn't store the photo.

PR #559, opened 2017-10-31
#549 Added choose and update button to google search users

jainanisha90:
- Cache master and resized image for chosen google search user
- Update candidate entry with newly cached images and google search link
- Mark chosen and updated entry with ok


DaleMcGrew:
👍

PR #570, opened 2017-11-02
Fixed #567 #549

### LINKED ISSUE #567: Move Google Search Results table to its own page
Show only the first row of the Google Search results table on the candidate edit page.

Please follow the pattern of putting the table in its own template file, so you can use the same exact table on two pages. Here is an example:
WeVoteServer/templates/analytics/sitewide_voter_metrics_table.html
...is included in:
WeVoteServer/templates/analytics/sitewide_voter_metrics.html

jainanisha90:
- Created google_search_users_for_candidate_table tamplate to load google serach users table.
- Show one google search user on candidate edit page with show more and show less link
- Choose and Update button for twitter search will not cache image instead refresh_candidate_twitter_details will do caching at the end of candidate_edit_process
- Updated likelihood_score for Ballotpedia, linkedin, facebook, twitter and wikipedia links
- Top google search result on candidate list page filters out on the basis of image, score
- Added Choose and update button to select google search user on candidate list page
- Added common function extract_website_from_url, extract_facebook_username_from_text_string

### DIFF
--- a/candidate/controllers.py
+++ b/candidate/controllers.py
@@ -19,7 +19,8 @@ from position.controllers import update_all_position_details_from_candidate
 from twitter.models import TwitterUserManager
 import requests
 import wevote_functions.admin
-from wevote_functions.functions import positive_value_exists, process_request_from_master, convert_to_int
+from wevote_functions.functions import positive_value_exists, process_request_from_master, convert_to_int, \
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
Great work Anisha!

PR #571, opened 2017-11-02
Fixed #567 Forgot to add google_search_users_for_candidate_table.html in previous checkin

### LINKED ISSUE #567: Move Google Search Results table to its own page

jainanisha90:


DaleMcGrew:
👍

Issue #568, opened 2017-11-01
Implement RemoteRequestHistory into retrieval of possible Twitter handles


edward-ly:
As a continuation from #538 and #544, we want to create/update a RemoteRequestHistory object each time we retrieve possible Twitter handles for a candidate or candidates. That way, we can also update the bulk retrieval process to know which candidates to skip since the retrieval has already been previously run, even if a candidate search returns zero results.

PR #574, opened 2017-11-02
Implemented RemoteRequestHistory into retrieval of possible Twitter handles, merged keywords lists

edward-ly:
Implemented RemoteRequestHistory into retrieval of possible Twitter handles to check whether the search has been done on a candidate before (#568). Fixed a bug preventing the status message parameter from being added to the RemoteRequestHistory object. Merged keywords lists from Google and Twitter searches into one common list.


DaleMcGrew:
👍

PR #585, opened 2017-11-08
Changes in Sign in with Facebook on admin page and Extract facebook user details from facebook when retrieve possible google search users.

jainanisha90:


DaleMcGrew:
👍

PR #587, opened 2017-11-09
Modifications in facebook search: Added facebook search details fields like facebook_about, facebook_description etc. to google custom search table

jainanisha90:


DaleMcGrew:
👍

PR #596, opened 2017-11-14
Minor changes to handle voter_device_id cookie for Native.

SailingSteve:
In views_organization/organization_retrieve_view and
views_voter/voter_issue_follow_view changeover to get_voter_device_id()
to retrieve the id from the request as needed by our Native Apps.
In import_export_twitter/controller new define TWITTER_NATIVE_INDICATOR
instead of using the string 'native' in two places in the code.


DaleMcGrew:
👍

Issue #115, opened 2016-07-09
Create organizational endorsements importer


DaleMcGrew:
We Vote is asking organizations to provide their endorsements in a JSON format. The first group to do this is Democracy for America, and they have provided this API endpoint:
http://www2.democracyforamerica.com/api/we_vote_usa_data.json

The specification for this format is here:
https://docs.google.com/document/d/1D90CslIjOPSxyB6XIBPSV07vDlVp5oMNyQPlOrq1Qj4/edit

TODO:
- [x] Add "organization_endorsements_api_url" URL field in the organization table to store the link to this JSON file for each organization. See: WeVoteServer/organization/models.py (Organization model)
- [x] Add a field on this edit page where this URL can be updated: http://localhost:8000/org/1201/edit/ See: WeVoteServer/organization/views_admin.py (organization_edit_view & organization_edit_process_view) and WeVoteServer/templates/organization/organization_edit.html
- [x] Add a link (similar to "Retrieve photos from Wikipedia" and "Scrape org website for Twitter Accounts") that reaches out to the API endpoint, and then stores the candidate and measure position data locally within the import_export_batches system as a new kind of batch set that will create candidates, measures, and the positions that go along with them.

udii:
A few questions:
1) the json includes candidate positions and measure positions, should we import both?
2) what field(s) should be used to look up the candidates/measures?
3) what should be done if a candidate/measure is not found? ignore it? create it? other??

DaleMcGrew:
Hi @udii, here are my thoughts:
1) the json includes candidate positions and measure positions, should we import both?
A: Yes please.

2) what field(s) should be used to look up the candidates/measures?
I would recommend these starting points for example code regarding bringing in:
a.) OFFICES & CANDIDATES: import_export_google_civic/controllers.py (process_contest_office_from_structured_json & process_candidates_from_structured_json)
b.) MEASURES/REFFERENDUM: import_export_google_civic/controllers.py (process_contest_referendum_from_structured_json)

3) what should be done if a candidate/measure is not found? ignore it? create it? other??
We should create new entries if not found. In candidate/models.py you can look for candidate_name or google_civic_candidate_name + state_code. You can also look for an existing record in our database by Twitter handle, or website. In measure/models.py it would be by measure_title + state code.

udii:
Without going into the full details, the update_or_create_\* methods we have in controllers typically require a google_civic_election_id and some we_vote_id to make an update.

[...]

PR #597, opened 2017-11-15
Create organizational endorsements importer #115

mnneelam:


DaleMcGrew:
Great code @mnneelam! 👍

PR #600, opened 2017-11-28
Started auth & retrieve tweets

eayoungs:
This barely works and is a total hack but it more proves the concept of what I'm trying to achieve.
- [ ] Is there a way to get the stored twitter credentials from the model rather than re-authorizing? (I poked around a bit but figured it's probably better to just ask)
- [x] The tweepy API function `user_timeline()` has a optional parameter for limiting the number of tweets returned but, as far as I can tell it's positional and I can't specifying it as a kwarg. I'll dig a little further into this later, as it's late right now so I'm probably overlooking something obvious.

### DIFF
--- a/organization/controllers.py
+++ b/organization/controllers.py
@@ -33,8 +33,11 @@ logger = wevote_functions.admin.get_logger(__name__)
 
 WE_VOTE_API_KEY = get_environment_variable("WE_VOTE_API_KEY")
 ORGANIZATIONS_SYNC_URL = get_environment_variable("ORGANIZATIONS_SYNC_URL")
+# Seems like the following lines are repeated from import_export_twitter/controllers.py
 TWITTER_CONSUMER_KEY = get_environment_variable("TWITTER_CONSUMER_KEY")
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Pull request approved! 👍 
Hi @eayoungs, regarding your question "Is there a way to get the stored twitter credentials from the model rather than re-authorizing? (I poked around a bit but figured it's probably better to just ask)"
You can look up the latest authorization session from the WeVoteServer/import_export_twitter/models.py TwitterAuthResponse table. There is a function in the TwitterAuthManager called "retrieve_twitter_auth_response" that returns the latest twitter_auth_response for the current voter_device_id.

bharathdn:
Thanks for the Review @DaleMcGrew :)

eayoungs:
Hey, that rimes! ;-)

Thanks, Dale!

eayoungs:
Hey, Dale; I'm having trouble finding where I can access the voter_device_id. It doesn't seem to be available in either `voter/models.py` or `organization/models.py`.

PR #606, opened 2017-12-01
retrieve_tweets_cached_locally

abdulrahimq:
- Cleaned code in organization/controllers.py
In Twitter/models.py:
- Added retrieve_tweets_cached_locallyin 
- Fixed is_retweet

I have a question about the OrganizationLinkToHashtag. Do we want to have a function in that class to create the organization_link_to_hashtag? Or do we want to set the attributes of the class instance in the organization_analyze_tweets? Setting the attributes looks weird to me but wants to be sure.

### DIFF
--- a/organization/controllers.py
+++ b/organization/controllers.py
@@ -55,6 +55,7 @@ def organization_retrieve_tweets_from_twitter(organization_we_vote_id, number_to
     twitter_user_manager = TwitterUserManager()
     tweets_saved = 0
     tweets_not_saved = 0
+
     for tweet_json in new_tweets:
         results = twitter_user_manager.update_or_create_tweet(tweet_json)
         if results['success']:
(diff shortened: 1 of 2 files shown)


organization/controllers.py
```diff
@@ -55,48 +55,54 @@ def organization_retrieve_tweets_from_twitter(organization_we_vote_id, number_to
...
-    for tweet in all_tweets:
+    for tweet in new_tweets:
         words = tweet.text.split()
         for word in words:
             if word in counts:
                 counts[word] += 1
             else:
                 counts[word] = 1
 
+    tweet_list = twitter_user_manager.retrieve_tweets_cached_locally()
```
DaleMcGrew, line 95: Hi @abdulrahimq, you will want to get "results" from twitter_user_manager.retrieve_tweets_cached_locally() and then get the tweet_list with:
tweet_list = results['tweet_list']

twitter/models.py
```diff
@@ -159,6 +159,10 @@ def update_or_create_tweet(self, tweet_json):
         """
 
         created = False
+        
+        is_retweet_boolen = False
```
DaleMcGrew, line 163: Spelling: is_retweet_boolean

organization/controllers.py
```diff
@@ -55,48 +55,54 @@ def organization_retrieve_tweets_from_twitter(organization_we_vote_id, number_to
...
     for tweet_json in new_tweets:
         results = twitter_user_manager.update_or_create_tweet(tweet_json)
         if results['success']:
             tweets_saved += 1
         else:
             tweets_not_saved += 1
 
-    # Gets ID, date and tweet text
-    # out_tweets = [[tweet.id_str, tweet.created_at, tweet.text] for tweet in new_tweets]
     return results
```
DaleMcGrew, line 66: Currently the results returned are the last results from update_or_create_tweet. I would build a results dict like this:
results = {
'success': True,
'status': status,  # If update_or_create_tweet returns a 'success' of False, you could capture the message above with status += update_or_create_tweet_results['status']
'tweets_saved': tweets_saved,
'tweets_not_saved': tweets_not_saved,
}

twitter/models.py
```diff
@@ -186,6 +192,33 @@ def update_or_create_tweet(self, tweet_json):
...
+        tweet_on_stage = Tweet()
+        tweet_found = False
+        success = False
+        status = "TWEET_NOT_FOUND"
+
+        tweet_list = []
+
+        try:
+            tweet_list_query = Tweet.objects.all()
+            tweet_list = list(tweet_list_query)
```
DaleMcGrew, line 206: I would recommend storing "organization_we_vote_id" with each Tweet, and then you can add a filter & sort like this:
tweet_list_query = tweet_list_query.filter(organization_we_vote_id__iexact=organization_we_vote_id)
tweet_list_query = tweet_list_query.order_by('date_published').reverse()

DaleMcGrew:
This is a good start Abed! 👍

PR #613, opened 2017-12-08
Updated Twitter handle stored in DB

eayoungs:
The **twitter model >** `update_or_create_tweet(...)` function was assigning the full organization's full name, rather than the Twitter handle. The full name was overflowing the max size in the database field. This was intermittent since many organizations' full name & handle are the same.


DaleMcGrew:
Good catch! 👍

PR #616, opened 2017-12-12
#1202 Email Ballot to Self: Voter name should not be used when "Voter-wv..."

jainanisha90:


DaleMcGrew:
Hi Anisha, I took a slightly different approach in pull request 617, where I am filtering out the "fake" names, like Voter-wv02voter111.

PR #622, opened 2017-12-13
Create organizational endorsements importer #115,

mnneelam:
added support for photo_url in BatchRowActionCandidate, support for storing candidate photo from organization endorsements in image table

### DIFF
--- a/candidate/models.py
+++ b/candidate/models.py
@@ -15,6 +15,7 @@ from wevote_functions.functions import convert_to_int, display_full_name_with_co
     extract_last_name_from_full_name, extract_suffix_from_full_name, extract_nickname_from_full_name, \
     extract_state_from_ocd_division_id, extract_twitter_handle_from_text_string, \
     positive_value_exists
+from image.models import ORGANIZATION_ENDORSEMENTS_IMAGE_NAME
 
 logger = wevote_functions.admin.get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


candidate/models.py
```diff
@@ -1376,6 +1379,11 @@ def create_candidate_row_entry(self, update_values):
                 new_candidate.candidate_twitter_handle = candidate_twitter_handle
                 new_candidate.candidate_url = candidate_url
                 new_candidate.facebook_url = facebook_url
+                new_candidate.photo_url = photo_url
+                if new_candidate.photo_url:
+                    # save_to_candidate_object = False
+                    candidate_results = self.modify_candidate_with_organization_endorsements_image(new_candidate, photo_url,
+                                                                                             False)
                 new_candidate.save()
```
DaleMcGrew, line 1387: I think before saving "new_candidate" we need to extract new_candidate from candidate_results that comes from "modify_candidate_with_organization_endorsements_image", yes?

candidate/models.py
```diff
@@ -1452,6 +1460,20 @@ def update_candidate_row_entry(self, candidate_we_vote_id, update_values):
...
+                        save_to_candidate_object = False
+                    else:
+                        save_to_candidate_object = True
+
+                    candidate_results = self.modify_candidate_with_organization_endorsements_image(
+                        existing_candidate_entry, update_values['photo_url'], save_to_candidate_object)
+                    if candidate_results['success']:
+                        values_changed = True
+
                 # now go ahead and save this entry (update)
```
DaleMcGrew, line 1477: I think you might want to get the updated "existing_candidate_entry" from the "candidate_results" here before the save a few lines down.

DaleMcGrew:
👍

mnneelam:
Thanks for pointing out the missed steps. I have updated code with these changes and also verified that code is working  (with a workable temporary candidate photo_url)

PR #627, opened 2017-12-19
Storing OrganizationLinkToHashtag; fails on iteration

eayoungs:
Using the frequency of hashtags as an indicator of an issue to be recommended to the organization (still needs to be matched with an actual WeVote issue to be followed). Currently only a single OrganizationLinkToHashtag object is stored. A `duplicate key value violates unique constraint` is thrown for further iterations, despite `setting=False` for fields and explicitly adding `id = models.AutoField(primary_key=True`.

### DIFF
--- a/organization/controllers.py
+++ b/organization/controllers.py
@@ -31,8 +31,6 @@ import re
 from organization.models import OrganizationLinkToHashtag
 from import_export_twitter.models import TwitterAuthManager
 
-import pandas as pd
-
 
 logger = wevote_functions.admin.get_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


organization/controllers.py
```diff
@@ -51,40 +50,56 @@ def organization_retrieve_tweets_from_twitter(organization_we_vote_id):
...
+    auth = tweepy.OAuthHandler(TWITTER_CONSUMER_KEY, TWITTER_CONSUMER_SECRET)
+    auth.set_access_token(TWITTER_ACCESS_TOKEN, TWITTER_ACCESS_TOKEN_SECRET)
+    api = tweepy.API(auth)
 
-        number_to_retrieve = 200  # TODO Remove this
-        organization_manager = OrganizationManager()
+    number_to_retrieve = 200  # TODO Remove this
+    organization_manager = OrganizationManager()
+    try:
         organization_twitter_id = organization_manager.fetch_twitter_handle_from_organization_we_vote_id(
```
DaleMcGrew, line 74: The try block should start before you call anything with Tweepy.

DaleMcGrew:
👍

eayoungs:
Thanks, Dale. Any thoughts on why the primary key isn't auto-incrementing in OrganizationLinkToHashtag() model?

PR #634, opened 2018-01-03
Fixing link to "Node/React/Flux WebApp"

robla:
Linking over to wevote/WebApp's "README_WORKING_WITH_WEB_APP.md" rather than to this doc


DaleMcGrew:
👍

Issue #583, opened 2017-11-04
If Twitter image has "_normal" in its name, we are unable to cache locally


DaleMcGrew:
This is an example:
https://api.wevoteusa.org/org/2531/pos/?google_civic_election_id=4315

We are not able to create local caches of this image:
twitter_profile_image_url_https: https://pbs.twimg.com/profile_images/1891380611/nyspffatwitlogo_normal.jpg

DaleMcGrew:
Fixed! Great work @mnneelam.

PR #638, opened 2018-01-04
#583, as per latest twitter photo size updates, original photo can is 400x400 size

mnneelam:
updated twitter_profile_image_url_https_original to reflect this photo size


DaleMcGrew:
👍

Issue #611, opened 2017-12-07
Merging Candidate Records


DaleMcGrew:
Candidates can be created in a few ways, and sometimes this will lead to duplicate candidate entries. We need administration tools that allow us to merge candidates quickly and in bulk, while preserving all of the gathered data.

More details here:
https://docs.google.com/document/d/1ezwrWo9pu5AVVHjFJHTdSMirNHF5a6QInjKJS6lMya0/edit#

DaleMcGrew:
Verified!

PR #647, opened 2018-01-06
Implemented Candidate Records Merge

edward-ly:
Implemented the first version for the process of merging two candidates, refactoring the candidate merge page and adding the ability to update key information for one candidate so that the other is ready to be deleted (#611).


position/controllers.py
```diff
@@ -226,6 +226,87 @@ def return_most_likely_data(from_position, to_position, attribute):
...
+def move_positions_to_another_candidate(from_candidate_id, from_candidate_we_vote_id,
+                                        to_candidate_id, to_candidate_we_vote_id,
+                                        public_or_private):
+    status = ''
+    success = True
+    position_entries_moved = 0
+    position_entries_not_moved = 0
+    position_list_manager = PositionListManager()
+
+    retrieve_public_positions = public_or_private
```
DaleMcGrew, line 238: This variable naming convention "retrieve_public_positions" is no longer valid and can be updated.

candidate/views_admin.py
```diff
@@ -903,19 +905,361 @@ def candidate_retrieve_photos_view(request, candidate_id):
...
+    redirect_to_candidate_list = request.POST.get('redirect_to_candidate_list', False)
+    remove_duplicate_process = request.POST.get('remove_duplicate_process', False)
+    state_code = request.POST.get('state_code', '')
+
+    candidate1_on_stage = {}
+    candidate2_on_stage = {}
+
+    candidate1_results = candidate_campaign_manager.retrieve_candidate_campaign_from_we_vote_id(candidate1_we_vote_id)
+    if candidate1_results['candidate_campaign_found']:
+        candidate1_on_stage = candidate1_results['candidate_campaign']
```
DaleMcGrew, line 933: I would create a variable "candidate1_id = candidate1_on_stage.id" here for use below. (and do the same for candidate2_id)

candidate/views_admin.py
```diff
@@ -903,19 +905,361 @@ def candidate_retrieve_photos_view(request, candidate_id):
...
+        if conflict_value == "CONFLICT":
+            choice = request.POST.get(attribute + '_choice', '')
+            if candidate2_we_vote_id == choice:
+                setattr(candidate1_on_stage, attribute, getattr(candidate2_on_stage, attribute))
+        elif conflict_value == "CANDIDATE2":
+            setattr(candidate1_on_stage, attribute, getattr(candidate2_on_stage, attribute))
+
+    # Merge public positions counts
+    public_positions_results = move_positions_to_another_candidate(candidate2_on_stage.id, candidate2_we_vote_id,
+                                                                   candidate1_on_stage.id, candidate1_we_vote_id,
```
DaleMcGrew, line 983: At this point in the code, I would use "candidate1_id" and "candidate1_we_vote_id" in favor of "candidate1_on_stage.id" (as opposed to one value from the incoming variable, and the other from a retrieved object) since the values in the candidate1_on_stage object will be in transition. You will need to create a variable candidate1_id as soon as the candidate1 object is retrieved.

DaleMcGrew:
Great work @edward-ly! 👍

Issue #637, opened 2018-01-04
Create Admin tool to update google_civic_election_id from a local version to official id


DaleMcGrew:
Google civic assigns an official id, called the google_civic_election_id, to every election they support. Unfortunately they don't always release this number until well after when We Vote needs to start preparing for an election. So we create elections with a number greater than 1,000,000 in place of this google_civic_election_id.

We need to create an admin tool that lets us change the google_civic_election_id used in several tables, from our internal number to the official one from Google.

To find all of the models that will probably need to be updated, search for:
google_civic_election_id = models.PositiveIntegerField
google_civic_election_id = models.CharField
 
This work has been started in "election_migration_view" WeVoteServer/election/views_admin.py
Go to an election page with a google_civic_election_id > 1000000 and you will see a "Migrate this Data into Google Civic Election" link which takes you to the start of this interface.

DaleMcGrew:
Done!

PR #658, opened 2018-01-18
#637 admin tool to update google_civic_election_id from a local version to official id

mnneelam:
added changes to election_migration_view to handle GET as well as POST request for migration. Added count_*_for_election function in candidates, contestoffice and position manager models for returning query count for a given google_civic_election_id.


DaleMcGrew:
👍 Thank you @mnneelam!

PR #670, opened 2018-02-03
Changes to handle twitter oAuth flows for Apache Cordova

SailingSteve:
In views_twitter.py, pass a 'cordova' boolean value through the various
steps, because if in iOS cordova, we need to use a native Apple
mechanism called a url scheme handler (wevotetwitterscheme) to allow us
to redirect back to the app.
In import_export_twitter/controllers, also pass around and add to URLs
the 'cordova' boolean value.
New template called cordova_ios_redirect_to_scheme.html, that contains
a one line JavaScript that sets the window.location.href with the Apple
custom scheme 'wevotetwitterscheme'


DaleMcGrew:
👍

PR #672, opened 2018-02-09
Issue 666

marodrig:
feature: Adds link to export csv file from batch list page.


DaleMcGrew:
Hi @marodrig, you seem to have checked in ~90 unnecessary files, starting with /apis_v1/static -- could you please remove all of those files from your pull request? There is also templates/500.html which looks like a test file.

Issue #654, opened 2018-01-14
Invite friends by email doesn't send invites entered before sender verifies email address


DaleMcGrew:
To test this, you will need two email accounts.
1. Make sure you are signed out
2. Go to https://wevote.us/friends/invitebyemail and enter the friend you want to add (ex/ test2@WeVote.info). Make sure to enter a first and last name.
3. Click send
4. Enter your own email address and name (ex/ test1@WeVote.info)
5. After you verify test1@WeVote.info, the email should be sent to test2@WeVote.info, but it is not sent.

Also, on the "Friend Requests" tab we can see that the first and last name did not get stored:
![screen shot 2018-01-14 at 7 41 00 am](https://user-images.githubusercontent.com/7756031/34917656-49e247c4-f8fe-11e7-8326-6beb6f57d0fa.png)

DaleMcGrew:
Verified!

PR #675, opened 2018-02-12
#654 Invite friends by email will send immediately send invites when sender verifies email address

jainanisha90:


DaleMcGrew:
Thank you @jainanisha90! 👍

PR #681, opened 2018-02-15
Saving Election data only if Ballot Returned data exists

jainanisha90:
Reverted election query in populate_election_data.py

### DIFF
--- a/ballot/models.py
+++ b/ballot/models.py
@@ -1377,6 +1377,25 @@ class BallotReturnedManager(models.Model):
             'ballot_returned':          ballot_returned,
         }
 
+    def should_election_search_data_be_saved(self, google_civic_election_id):
+        success = False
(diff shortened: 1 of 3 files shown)


ballot/models.py
```diff
@@ -1377,6 +1377,25 @@ def find_closest_ballot_returned(self, text_for_map_search, google_civic_electio
...
+            if ballot_returned_list_count > 0:
+                success = True
+                status = 'ELECTION SEARCH DATA BE SAVED'
+
+        results = {
+            'success':  success,
+            'status':   status,
+        }
+        return results
+
```
DaleMcGrew, line 1398: Hi Anisha, for a simple function like "should_election_search_data_be_saved" we don't need to  return results. Instead we can just return True or False.

Issue #695, opened 2018-03-01
Refresh from Twitter: Problem with 200x200 images


DaleMcGrew:
From an organizational position page:
https://api.wevoteusa.org/org/2741/pos/

...when I use "Refresh from Twitter", the largest image, which should be 200x200, is only retrieved as 48x48: https://api.wevoteusa.org/import_export_twitter/2741/refresh_twitter_organization_details/

I think this may be related to the recent removal of the use of "_normal" attached to some image names by Twitter.

![screen shot 2018-03-06 at 10 22 26 am](https://user-images.githubusercontent.com/7756031/37050303-5ca76d7e-2128-11e8-9e0d-7f339a972f9b.png)

DaleMcGrew:
@mnneelam this is an example of one of the images stored at much lower resolution: https://wevote-images.s3.amazonaws.com/wv02voter102240/twitter_profile_image-20180105_1_200x200.jpeg

DaleMcGrew:
Fixed!

PR #700, opened 2018-03-06
Fix for #695 using original instead of _normal for twitter_profile_image

mnneelam:


DaleMcGrew:
👍  Instead of showing the large image, I changed the HTML to be a link to open the image in a new window.

PR #701, opened 2018-03-07
Fixed a bug when multiple entries existed for images with same day but different versions. It picks the latest saved image from the list

mnneelam:


DaleMcGrew:
👍

PR #724, opened 2018-04-12
Export voters emails

marodrig:
Created a method that creates a batch from a list of objects. Currently it assumes the Class of the object to be Voter.  Future to-do will include making this method more flexible.


DaleMcGrew:
👍

PR #732, opened 2018-04-16
Handle login by email for Cordova, iOS only

SailingSteve:
Added a is_cordova boolean to use the redirect scheme if Cordova
Needed for the fix to the iOS part of wevote/WeVoteCordova/issues/11
Android will need a separate approach


DaleMcGrew:
👍

PR #851, opened 2018-07-16
Keep the load balancer from rotating to another node during a long query

SailingSteve:
New JavaScript in election_list that hits an API every two seconds to
avoid having the load balancer session timeout.

"Fixes" wevote/WeVoteServer/issues/822


DaleMcGrew:
👍 Thank you @SailingSteve!

Issue #840, opened 2018-07-11
Retrieve candidate Facebook "page" photo


DaleMcGrew:
Candidates will list a Facebook page with their public profile. We would like to be able to retrieve this photo and use it for the candidate.

Can we use the Google Custom Search Engine tool we have to retrieve photos from Facebook pages?

See: https://developers.facebook.com/docs/graph-api/reference/page/picture/

SailingSteve:
```<a aria-label="Profile picture" class="_2dgj" href="/24413227922/photos/10155764935092923/" rel="theater"><img class="_4jhq img" height="180" src="https://scontent.fsnc1-1.fna.fbcdn.net/v/t1.0-1/p480x480/19420547_10155764935092923_4741672932915645516_n.jpg?_nc_cat=0&amp;oh=2cac9b689b5117adf09ecae347fbe06b&amp;oe=5BDE0687" width="180" alt=""></a>```

These would be pretty easy to scrape, if we needed to
https://scontent.fsnc1-1.fna.fbcdn.net/v/t1.0-1/p480x480/19420547_10155764935092923_4741672932915645516_n.jpg?_nc_cat=0&oh=2cac9b689b5117adf09ecae347fbe06b&oe=5BDE0687

<a aria-label="Profile picture" class="_2dgj" href="/24413227922/photos/10155764935092923/" rel="theater"><img class="_4jhq img" height="180" src="https://scontent.fsnc1-1.fna.fbcdn.net/v/t1.0-1/p480x480/19420547_10155764935092923_4741672932915645516_n.jpg?_nc_cat=0&amp;oh=2cac9b689b5117adf09ecae347fbe06b&amp;oe=5BDE0687" width="180" alt=""></a>

DaleMcGrew:
The challenge is starting with a Facebook page (Ex/ https://www.facebook.com/KamalaHarris/ ) and then getting back a link to this profile photo) We could also do a CURL call and just look for the URL in the HTML? It would be better if we could do it through the Facebook API thought.

Once we have the image URL, we have a whole system for taking it in, resizing it to various sizes, and caching it on our AWS servers.

SailingSteve:
@DaleMcGrew 
It looks like we would need a page access token to get the picture through the API, in a similar way to the way we get their banner (I made that change last year).  So they would have to be "logged in" to WeVote through facebook.  

I've done lots of scraping in my past, this is an easy one.

[...]

PR #862, opened 2018-07-20
Fix facebook scrape regex to support a second image url markup scheme 

SailingSteve:
Fix to wevote/WeVoteServer#840
"Tim Gildersleeve" and "Cort VanOstran" have a sightly different markup
not the regex supports their rare layout.


DaleMcGrew:
👍 Thank you!

PR #866, opened 2018-07-23
Twitter retrieve should save candidate website

SailingSteve:
wevote/WeVoteServer/issues/819 
In refresh_twitter_candidate_details_for_election If we have a twitter_url, but no candidate_twitter_handle, extract the handle and save it. 
Then if we have a candidate_twitter_handle, but no twitter_url, build a twitter_url and save it. 
Then if we have a twitter_url, but no candidate_url, save the twitter_url as a candidate_url


DaleMcGrew:
👍

PR #903, opened 2018-08-27
Resolve runtime exception in update_or_create_ballot_returned()

SailingSteve:
Was only looking to avoid "False", but was receiving "None" in election_day_text


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #907, opened 2018-08-31
If field is missing, sort it last.

josephevans:


DaleMcGrew:
👍 Thank you @josephevans.

PR #926, opened 2018-09-12
Better error message when you can't get a Facebook candidate picture via the supplied URL.

SailingSteve:
Improves error message from wevote/WeVoteServer/issues/869


DaleMcGrew:
👍

PR #935, opened 2018-09-24
Retrieve Elected Officials from Google's representativeInfoByAddress API

SailingSteve:
wevote/WeVoteServer/issues/768
This feature is launched from the Offices tab, and loads the appropriate
info for Officials at the same time -- they are in the same dataset, and
retrieving them both at the same time keeps them in sync, with a
potentially changing database at Google.


DaleMcGrew:
Your real time status code looks very interesting. Thank you @SailingSteve! 👍

PR #955, opened 2018-10-12
updating README to include Homebrew installation instructions for Pos…

amva13:
updating README to include Homebrew installation instructions for PostgreSQL. I found this an easier approach than the traditional dmg drag & drop / change paths (may be true for others who work more on command-line as well).


DaleMcGrew:
👍 Thank you @amva13 for improving our set up documentation!

Issue #798, opened 2018-06-17
Please add Ballotpedia API calls to Quick Statistics page


DaleMcGrew:
On this page, we show recent calls to the various API's we use:
https://api.wevoteusa.org/admin/statistics/

The view that returns this data is "statistics_summary_view" in WeVoteServer/admin_tools/views.py

Please add a statistics counter that totals up all of the number of Ballotpedia API calls.

PR #974, opened 2018-11-02
Fix for https://github.com/wevote/WeVoteServer/issues/798

unsalc:
Fixed https://github.com/wevote/WeVoteServer/issues/798


DaleMcGrew:
👍 Thank you for your first We Vote check in @unsalc!

PR #977, opened 2018-11-02
Strip html tags

asreepathy:
Using django.util.html.strip_tags to strip html tags.
Issue - https://github.com/wevote/WebApp/issues/1788


DaleMcGrew:
👍 Thank you @asreepathy for your first pull request with We Vote!

PR #983, opened 2018-11-10
Added a remove facebook authentication button to the admin voter page

SailingSteve:
Tested with test data from a sql file, real test will be when it goes on
the live server (due to facebook origin restrictions).
Added favicons to the admin app


DaleMcGrew:
👍

SailingSteve:
Tested on the live server, works as expected.

PR #992, opened 2019-01-08
Minor change that should avoid a http 400 on Stripe /v1/plans

SailingSteve:
A field had been renamed in the latest API
Please send me a note when this makes it to the production server.  Thanks


DaleMcGrew:
👍

PR #993, opened 2019-01-08
Second change that should avoid a http 400 on Stripe /v1/plans

SailingSteve:
A dummy product field is now required.
No idea why this did not show up in the testing I did with the test api.


DaleMcGrew:
👍

DaleMcGrew:
Hello @SailingSteve, I have updated the live server with these changes.

PR #995, opened 2019-01-13
Switch Maxmind GeoIP from legacy version to GeoIP2 database format

SailingSteve:
Resolves wevote/WeVoteServer/issues/994

### DIFF
--- a/config/environment_variables-template.json
+++ b/config/environment_variables-template.json
@@ -154,5 +154,8 @@
   "ISSUES_IMAGE_LARGE_WIDTH":       200,
   "ISSUES_IMAGE_LARGE_HEIGHT":      200,
 
+  "_comment":                       "GeoLite2 IP address database location",
+  "GEOLITE2_DATABASE_LOCATION":     "geoip2/city-db/GeoLite2-City.mmdb",
+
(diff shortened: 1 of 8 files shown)


SailingSteve:
'GEOLITE2_DATABASE_LOCATION' works on my local pointing to the db with a relative and a full path, so it should be ready for Joe.

"geoip2/city-db/GeoLite2-City.mmdb"
"/Users/stevepodell/PythonProjects/WeVoteServer/geoip2/city-db/GeoLite2-City.mmdb",

SailingSteve:
@DaleMcGrew Could you merge this, it would make my work on the Django upgrade easier.  Thanks, Steve

DaleMcGrew:
Hi @SailingSteve I will merge this now, although this pull request will complicate my ability to update the live server today. To make a long story short (I sent you an email about this), please make sure to put all of your Django upgrades in a new branch (as opposed to `develop`), because we will need to thoroughly test the WeVoteServer Django upgrade (which may take 1-2 weeks) before we push it to the develop branch. I would like to make sure we can keep making updates to the develop branch (and thus the live server) while the Django upgrade branch is being tested.

PR #1015, opened 2019-03-25
Added a Django Admin page that shows the distribution of polling places on a Google Map.

SailingSteve:
Added a "Google Maps Polling Locations" page that shows the geographic distribution of polling places.
Use as a diagnostic to see if we have gaps in coverage.
As of today, look at Florida and zoom out on Alaska to see data issues.
Uses the Google Maps JavaScript API.  Location popup with address added.
Note:  The Maps API token is encoded in the templates/polling_location/polling_location_visualize.html file.


DaleMcGrew:
👍

PR #1048, opened 2019-07-23
Upgrading facebook-api library from v2.0 to v3.1

SailingSteve:
The latest Facebook SDK on their API servers is 3.3, but a corresponding
Python facebook-api library is not yet in general release.


DaleMcGrew:
👍

PR #1052, opened 2019-07-24
added some notes about the specifics of my installation

irascible:
Added some notes in the form of comments in the respective readme files. Most of the installation process was smooth, with small differences due to my existing mac installations of both python 2 and python 3. The one fairly important (but easy to fix) glitch was that I had to install postgres before pip install -r requirements.txt would complete because the former installs pg_config which the latter depends on.


DaleMcGrew:
Thank you @irascible for helping to improve the installation instructions! Approved. 👍

PR #1053, opened 2019-07-25
Simplified/refactored donation processing with Stripe

SailingSteve:
Added a bitmap field for business subscription feature enabling.


DaleMcGrew:
👍

Issue #1049, opened 2019-07-23
Fix broken link when suggested voter guide links are added without "http.."


DaleMcGrew:
An example suggested voter guide: https://api.wevoteusa.org/vg/create/?voter_guide_possibility_id=1503

(If you sign in with Twitter or Facebook, and then Slack me your We Vote Id, I can give you access to this page.)

"Link to voter guide" just uses the raw link, so we should probably tack on an "http://" for links that don't have "http" at the start of the url.
![Screen Shot 2019-07-23 at 1 34 59 PM](https://user-images.githubusercontent.com/7756031/61745977-8b252600-ad4f-11e9-895d-45da4dbcae30.png)

To navigate the interface:
1. Go to the Voter Guides section
2. Click on "Review Suggested Voter Guides"
3. Click on "Suggest a Voter Guide" and paste in "www.lidbk.com/2019-endorsements" as an example. "Analyze & Save"
4. Then click on the "Edit" button
5. Click on the right-arrow to the right of "Link to voter guide"

The template is here: WeVoteServer/templates/voter_guide/voter_guide_create.html
The view that prepares data for this page is "voter_guide_create_view" in the file WeVoteServer/voter_guide/views_admin.py

irascible:
it's possible that some pages might need an 'https' rather than an 'http'.

DaleMcGrew:
Verified! Thank you for your first check in with We Vote @irascible!

PR #1054, opened 2019-07-26
prefix link with http or https if necessary

irascible:
Address https://github.com/wevote/WeVoteServer/issues/1049 by prefixing a url with http or https as necessary. The prefixing occurs after the voter guide link page has been created; another way to go might be to add the prefix when the voter guide link page is being created.

Tried to keep within the general stylistic approach of existing code, but please let me know if not.  Also removing the 'href' from the link changed the color of the glyph to black and I hard-coded it back to blue--not sure that was kosher.

### DIFF
--- a/templates/voter_guide/voter_guide_create.html
+++ b/templates/voter_guide/voter_guide_create.html
@@ -71,9 +71,35 @@
     <label for="voter_guide_possibility_url_id" class="col-sm-3 control-label">
     {% if voter_guide_possibility_url %}
         Link to voter guide&nbsp;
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
This looks great @irascible! Thank you for your first We Vote pull request. We could definitely add the "http" or "https" upon save of the URL, and that might be a good next project. I'll write up an issue for that and assign it to you. This pull request is still important since we can't always be sure that the links that make it into our database will start with "http"

The thumbs up means that I code-reviewed this: 👍

Issue #1055, opened 2019-07-26
Fix link dynamically on Suggested Voter Guides page: when suggested voter guide links are added without "http.."


DaleMcGrew:
Fix the equivalent error like this one ( https://github.com/wevote/WeVoteServer/issues/1049 ) except on the Possibility List page, under "Voter Guide Source": http://localhost:8000/vg/possibility_list/?google_civic_election_id=0

![Screen Shot 2019-07-26 at 3 15 14 PM](https://user-images.githubusercontent.com/7756031/61984203-32a49180-afb8-11e9-83c9-a7fd09ffb3ab.png)

DaleMcGrew:
Verified. Thank you @irascible!

PR #1059, opened 2019-07-30
issue 1055: use template to dynamically prefix http to link

irascible:
Address https://github.com/wevote/WeVoteServer/issues/1055 by moving code from https://github.com/wevote/WeVoteServer/issues/1049 into a template file templates/voter_guide/voter_guide_dynamically_prefix_link_protocol.html


irascible:
Note: in the initial commit (fa35872) the <script> section would be reloaded every time the template is used (e.g. for every entry in the table on the Suggested Voter Guides page). In the second commit (e0f516d) the <script> is loaded only once by adding a 1-based counter variable to the template. It may be that approach is over-complicated, and it would be better to put the <script> section in a separate template to be loaded once per page.

DaleMcGrew:
Thank you @irascible! 👍

Issue #1062, opened 2019-08-01
Suggest Voter Guides List: Multi-select form to change VoterGuidePossibility status


DaleMcGrew:
On the Suggested Voter Guides page: https://api.wevoteusa.org/vg/possibility_list/?google_civic_election_id=1000080
...we need a way to select multiple rows (with check boxes) and then perform one of the following actions "Mark as...":

- Endorsements Not Available Yet (field: cannot_find_endorsements)
- Capture Detailed Comments (field: capture_detailed_comments)
- Candidates/Measures Missing (field: candidates_missing_from_we_vote)
- Stop Reviewing (field: hide_from_active_review)
- Ignore this Website (field: ignore_this_source)

These are currently set on the page where you are editing each Suggested Voter Guide, like this one:
https://api.wevoteusa.org/vg/create/?voter_guide_possibility_id=1508
![Screen Shot 2019-08-01 at 8 55 38 AM](https://user-images.githubusercontent.com/7756031/62308422-29914580-b43a-11e9-9ae9-e1f8345bd2fb.png)
And the data is stored in `VoterGuidePossibility` in `WeVoteServer/voter_guide/models.py`

Before starting, review this "stack" to see how the url and view take in input:
`views_admin.voter_guide_possibility_list_view` in `WeVoteServer/voter_guide/urls.py`
`voter_guide_possibility_list_view` in `/WeVoteServer/voter_guide/views_admin.py`

You will need to add a form to this page:
WeVoteServer/templates/voter_guide/voter_guide_possibility_list.html
...and the checkboxes to this included template page:
WeVoteServer/templates/voter_guide/voter_guide_possibility_table.html

You will need to set up the following files to process the form:
`views_admin.voter_guide_possibility_list_process_view` in `WeVoteServer/voter_guide/urls.py`
`voter_guide_possibility_list_process_view` in `/WeVoteServer/voter_guide/views_admin.py`

After processing `voter_guide_possibility_list_process_view`, return the browser to `https://api.wevoteusa.org/vg/possibility_list/`:
`voter_guide_possibility_list_view` in `WeVoteServer/voter_guide/views_admin.py`using code like this: `return HttpResponseRedirect(reverse('voter_guide:voter_guide_possibility_list', args=()))`

[...]

PR #1063, opened 2019-08-05
Suggest Voter Guides List: Multi-select form to change VoterGuidePossibility status

irascible:
https://github.com/wevote/WeVoteServer/issues/1062

Functionality seems to be working, but there are a couple of known formatting issues so far: label overlaps `<select>` and the new "Show 'Ignore this Website'" filter string is not evenly spaced with the rest:

Screen Shot 2019-08-04 at 4 48 34 PM

As ever, happy to hear comments on coding style, etc.


DaleMcGrew:
This looks great @irascible! Excellent work.

PR #1066, opened 2019-08-07
Admin display

Farid-Amirli:


templates/organization/organization_position_list.html
```diff
@@ -37,8 +55,16 @@ <h1>{{ organization.organization_name }}: Position List</h1>
                class="u-no-break"><span class="glyphicon glyphicon-new-window"></span></a>
             {% endif %}
         </td>
+         {% endif %}
```
DaleMcGrew: It looks like you have an extra `endif` here? I would recommend removing the `if organization.organization_website` and both `endif` statements.

templates/organization/organization_position_list.html
```diff
@@ -62,51 +88,87 @@ <h1>{{ organization.organization_name }}: Position List</h1>
...
+        <td>
+            {{organization.chosen_html_verification_string|default_if_none:""}}
+        </td>
+
     </tr>
-    {% endif %}
-    {% if organization.twitter_description %}
+
     <tr>
+        {% if organization.twitter_description %}
```
DaleMcGrew: I would remove this `if organization.twitter_description` (and the corresponding `endif`) so a missing twitter_description doesn't shift the Logo URL into the left column.

templates/organization/organization_position_list.html
```diff
@@ -62,51 +88,87 @@ <h1>{{ organization.organization_name }}: Position List</h1>
...
+        {% endif %}
+        <th> LOGO URL:</th>
+        <td>
+            {{organization.chosen_logo_url_https|default_if_none:""}}
+        </td>
     </tr>
-    {% endif %}
-    {% if organization.twitter_location %}
     <tr>
+        {% if organization.twitter_location %}
```
DaleMcGrew: I would remove this `if organization.twitter_location (and the corresponding endif) so a missing twitter_location doesn't shift the "Social share description" into the left column.

templates/organization/organization_position_list.html
```diff
@@ -62,51 +88,87 @@ <h1>{{ organization.organization_name }}: Position List</h1>
...
+        <th> Social share image:</th>
+        <td>
+            {{organization.chosen_social_share_image_256x256_url_https|default_if_none:""}}
+        </td>
+
     </tr>
-    {% endif %}
-    {% if organization.organization_email %}
     <tr>
+       {% if organization.organization_email %}
```
DaleMcGrew: I would remove this `if organization.organization_email` (and the corresponding endif) so a missing organization_email doesn't shift the "Sub domain string" into the left column.

DaleMcGrew:
Hello @FaridAmirli1990 I made some comments. Could you please let me know when you address them? One of them was a problem that I think would prevent the page from running. Thank you!

DaleMcGrew:
Thank you @FaridAmirli1990! 👍

PR #1072, opened 2019-08-14
Intermediate checkin for business plans with Stripe

SailingSteve:
Validation of coupons works.
Plumbing to get extra data for business subscriptions so we can reuse
the DonationPlanDefinition object is done.
Donation Subscriptions re-tested, to make sure they still work.


DaleMcGrew:
👍

PR #1073, opened 2019-08-16
Intermediate checkin for organization paid plans with Stripe

SailingSteve:
Requires a db migration

Validation of coupons now sets a few default coupons, mostly for developers.
Changed BusinessSubscriptionPlan to OrganizationSubscriptionPlan to
distinguish from donation subscription plans which use the same api calls
and mostly the same tables.


DaleMcGrew:
👍

PR #1081, opened 2019-08-23
New way to handle monthly subscriptions

SailingSteve:
Dropped and re-initialized the OrganizationSubscriptionPlans table
Added an annualized calculated field on the Coupon Details dialog.

Requires a migration


DaleMcGrew:
👍

PR #1089, opened 2019-08-29
Subscription to a paid plan is working, but still needs edge case handling work

SailingSteve:
Still need to prevent two paid plans for one organization
Unsubscribe from a paid plan needs work.

Work towards the completion of wevote/WebApp/issues/2488


DaleMcGrew:
👍

PR #1092, opened 2019-09-04
retrieve_voter_guide_possibility api now returns last URL match instead of first

SailingSteve:
Needed for Chrome Endorsement Extension


DaleMcGrew:
👍

PR #1114, opened 2019-10-03
First version of voterVerifySecretCode. Updated VoterDeviceLink to include new secret code system. Now sending emails with secret code - sign_in_code_email.

DaleMcGrew:
First version of voterVerifySecretCode. Updated VoterDeviceLink to include new secret code system. Now sending emails with secret code - sign_in_code_email.


SailingSteve (commented):
@DaleMcGrew 
Removing the wevotetwitterscheme here breaks the feature for existing released WeVoteCordova apps.

PR #1154, opened 2019-11-24
Cleared some travis warnings, experiment to upgrade node for travis runs.

SailingSteve:
The experiment is the travis.yml content -- I think that is all we need.


DaleMcGrew:
👍

PR #1160, opened 2019-12-03
Delete Images on /voter/ removes cached versions then blanks image fields for Voter, Twitter, and Facebook

SailingSteve:
To support testing of wevote/WeVoteServer/issues/1158


DaleMcGrew:
👍

PR #1163, opened 2019-12-07
Remove donation_history from voterRetrieve, that data is only needed in the DonationStore

SailingSteve:
The donationHistory api call is what provides that data now.


DaleMcGrew:
👍

PR #1186, opened 2020-01-17
Version upgrades to address Severe Alerts from Snyk. Had to comment out background tasks temporarily until we can look into this (labeled with comments "# DALE 2020-01-17 To prevent ModuleNotFoundError")

DaleMcGrew:
Version upgrades to address Severe Alerts from Snyk. Had to comment out background tasks temporarily until we can look into this (labeled with comments "# DALE 2020-01-17 To prevent ModuleNotFoundError")


DaleMcGrew:
Hi @SailingSteve, when you get a moment can you review this pull request? I have tested locally and none of these upgrades affect anything as far as I can tell so far.

PR #1188, opened 2020-01-20
Uncommented  and fixed scheduled tasks

SailingSteve:
There was a breaking change in Django background tasks 1.2.5, so the prior version upgrade broke things.


DaleMcGrew:
👍

PR #1239, opened 2020-03-07
voterGuidePossibilityHighlightsRetrieve can now receive a google election id

SailingSteve:
This is needed in order to continue testing with a prior election.
I don't have the data on my local to confirm that this works, but the id value gets through to the internal method call that already had accepted an id, if that method works, the updated API call should work.


DaleMcGrew:
👍

DaleMcGrew:
This is now on the live API server @SailingSteve.

PR #1252, opened 2020-03-26
new API pdfToHtmlRetrieve to convert PDFs in the Chrome Extension to HTML files in S3

SailingSteve:
Reads in the PDF by URL, applies it to pdf2Text.py by calling the Python tool https://github.com/pdfminer/pdfminer.six  (There are many other forks of pdfminer in Github, but the one we use is pdfminer.six, not plain pdfminer!)
Files in s3 are set to expire in 1 week
A good test URL:
https://www.iuoe399.org/media/filer_public/45/77/457700c9-dd70-4cfc-be49-a81cb3fba0a6/2020_lu399_primary_endorsement.pdf
Becomes:
https://wevote-temporary.s3.amazonaws.com/tmprts8w5ub.html

### DIFF
--- a/apis_v1/documentation_source/pdf_to_html_doc.py
+++ b/apis_v1/documentation_source/pdf_to_html_doc.py
@@ -1,0 +1,62 @@
+# apis_v1/documentation_source/pdf_to_html_doc.py
+# Brought to you by We Vote. Be good.
+# -*- coding: UTF-8 -*-
+
+
+def pdf_to_html_retrieve_view(url_root):
+    """
+    Show documentation about pdfToHtmlRetrieve
+    """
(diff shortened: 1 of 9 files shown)


DaleMcGrew:
Thank you @SailingSteve!  I'm wondering if we can add "pdfminer.six" to our WeVoteServer/requirements.txt file so that our production servers just pick it up instead of having to add it manually?

@josephevans do you have any thoughts about automating Steve's instructions (in this pull request) re: creating a symlink on the live servers? 
```Then in a terminal window, create a symlink to the executable script pdf2txt.py```

DaleMcGrew:
👍

PR #1324, opened 2020-05-23
Tests for email_address_save endpoint

Eduardo445:
We wrote tests for the following cases:
1) Request does not include a voter_device_id
2) Make sure email address is valid
3) If all conditions are met, then email address should save successfully


DaleMcGrew:
Thank you @Eduardo445! 👍

PR #1335, opened 2020-05-29
Added APIs and a table for Sign in with Apple

SailingSteve:
Sign in with Apple, is initially only for iOS under Cordova (To meet Apple's app submission requirement)
Added a doc page for appleSignInSave
Added comments in classes that unnecessarily extend model.Models, thereby forcing Django ORM to create a useless empty table with only an ID field.
In retrieve_organization_reserved_hostname() return 'hostname_is_reserved': True, for localhost


DaleMcGrew:
👍  Reviewed and approved.

PR #1371, opened 2020-07-06
Try removing AppleOAuth2 from list of AUTHENTICATION_BACKENDS

SailingSteve:
to avoid twitter/facebook signin to api server crash


SailingSteve:
I don't think I need the addition of the backend on  that list, it would only be if we wanted Login with Apple to the login to the API server.

Then it would be 'WeVoteServer.apple.AppleOAuth2' not just 'AppleOAuth2'

PR #1392, opened 2020-08-06
Sign in with Apple complete

SailingSteve:
Still need to test Apple on the live server
Cleared a bunch of: SyntaxWarning: "is" with a literal. for Python 3.8+:

- "The compiler now produces a SyntaxWarning when identity checks (is and is not) are used with certain types of literals (e.g. strings, numbers). These can often work by accident in CPython, but are not guaranteed by the language spec."


DaleMcGrew:
👍

PR #1394, opened 2020-08-07
Fix to Sign in with Apple redirect after decrypting the JWT

SailingSteve:
Works on my local, now we'll see on the live server


DaleMcGrew:
👍

PR #1396, opened 2020-08-10
Now capture the first last name on first query (that's the only chance we have) for Sign in with Apple

SailingSteve:
One line in AppleUser per Apple `user_code`, we no longer store the device_id in this table


DaleMcGrew:
👍

PR #1412, opened 2020-08-19
Implemented first draft of account merging with Apple Sign in. Still needs to be tested.

DaleMcGrew:
Implemented first draft of account merging with Apple Sign in. Still needs to be tested.


SailingSteve (approved):
This looks good!

PR #1436, opened 2020-09-08
Setup for sending a Firebase Cloud Message to a single phone, updates  the badge number

SailingSteve:
iOS only for now.
Requires a private key file to be stored on the server, outside of the Python app


DaleMcGrew:
👍

PR #1439, opened 2020-09-11
Firebase Cloud Message now works with Android

SailingSteve:
Added a field to VoterDeviceLink, that stores platform type, one of "IOS", "ANDROID", "WEBAPP"


DaleMcGrew:
👍

PR #1455, opened 2020-09-19
Firebase Cloud Message sending, runs in a thread, initiated from ActivityPostSave

SailingSteve:
Runs the cloud notification in a thread, so that the call to ActivityPostSave is not slowed down by the extra work.

This uses the Python version of the Java threads API, but CPython is inherently single threaded -- it runs only one 
thread at a time, and switches execution between threads to simulate simultaneous execution.  CPython does not provide the concurrent threading that you get in Java or C, where threads can be started up to run concurrently on other processor cores, but it does allow activity_post_save_view(request) to return quickly (in about 70ms on a MacBookPro), and then runs the thread (as currently loaded down with lots of temporary logging) in about 300ms for a single fcm notification send.  

(As an experiment I ran firebase_notification_send_to_cordova_apps 20 times in a loop within the thread and the time to execute activity_post_save_view was unchanged, but the thread took about 2 seconds to complete.)  

So with our load balancing setup, a Django instance would be tied up for an average of about 70ms for the ActivityPostSave execution, and would continue to be further tied up with the 300ms for the "notification send" thread to run to completion. 

This eliminates the need to batch process the notifications.


SailingSteve:
Partially addresses wevote/WeVoteServer/issues/1450
Still needs some testing when this makes it to the live server.

DaleMcGrew:
👍

PR #1472, opened 2020-10-20
Clarify which database fields in environment_variables.json need to be configured

ecluster:
While setting up the codebase on a Linux environment, I ran into an error that was ultimately caused by a few missing configurations in environment_variables.json. It took me over an hour to debug this. I figured I would add a quick clarification in the documentation so that other people do not make the same mistake.


DaleMcGrew:
Thank you Edward! I appreciate this clarification. Please let me know if you are ever interested in picking up any development projects.

ecluster:
Hi Dale,

No problem with the clarification of the documentation. I will
definitely keep in mind for any interest in other development projects! I
am happy that I can contribute to this open-source!

PR #1478, opened 2020-10-28
Tests for the email_address_retrieve endpoint

aundrelab:
We tested for the following cases

1) Pass no voter_device_id
2) Pass a valid voter_device_id, but user has no email address
3) Pass a valid voter_device_id and the user has a valid email address


DaleMcGrew:
Thank you @aundrelab! 👍

PR #1486, opened 2020-11-12
New retrieveSQLTables API endpoint to return the tables as CSV, that would otherwise be "Sync"ed

SailingSteve:
This is the "send" csv files half of replacing/supplementing the Sync with master feature


DaleMcGrew:
👍

PR #1488, opened 2020-11-15
New retrieveSQLTables API endpoint uses | as a delimiter and now exports a header line

SailingSteve:
Since tables created from scratch will have different column ordering than those migrated incrementally


DaleMcGrew:
👍

PR #1493, opened 2020-11-24
New retrieveSQLTables API endpoint -- reduce server memory use

SailingSteve:
by breaking apart the api, into a table by table query


DaleMcGrew:
👍

PR #1495, opened 2020-11-25
New retrieveSQLTables API endpoint -- add ability to break the 7M row ballot_ballotitem into chucks

SailingSteve:


DaleMcGrew:
👍

PR #1496, opened 2020-11-26
New retrieveSQLTables API endpoint -- extract large tables by range to avoid out of memory crash in server

SailingSteve:
Out of memory seems to fail silently, and does not have any reported error in Splunk.
This seems probable since the caught error in a try/except block in this code returns without the caught error appended to the output string.


DaleMcGrew:
👍

PR #1498, opened 2020-11-27
New retrieveSQLTables API endpoint -- working

SailingSteve:
Completed
It worked on my local at 45 seconds elapsed time, but it used gigabytes of memory and the production server could not handle it.  So now I pull in a million rows at a time, and it takes about 12 minutes.  Still very fast, and it brings in all election data for all states.
Bumped Cryptography version to 3.2   (used in notifications)


DaleMcGrew:
👍

PR #1526, opened 2021-01-15
Added a graph of voters with more than 3 friends on the /admin page

SailingSteve:
It is fast, but it still is only generated when you press the "Show Chart" button.


DaleMcGrew:
👍

PR #1528, opened 2021-01-15
Integrated PDF scanner into VG Possibility Scanner

SailingSteve:
The [http://localhost:8000/apis/v1/docs/pdfToHtmlRetrieve/](http://localhost:8000/apis/v1/docs/pdfToHtmlRetrieve/ ) api still works after the change.
When I override possibilites_found in voter_guide/views_admin.py at line 898, the code is
invoked and creates perfect looking output like
https://wevote-temporary.s3.amazonaws.com/Endorsements2020G.html
and at line 2033 in candidate/controllers.py the new html page is substituted for the PDF and that
should work, but I don't have data in my local to test it, or know how I would do it on the live server.
Now reports Node version running on the EC2 instance at bottom of the /admin/ page.

Should address wevote/WeVoteServer/issues/1254

### DIFF
--- a/admin_tools/views.py
+++ b/admin_tools/views.py
@@ -2,7 +2,7 @@
 # Brought to you by We Vote. Be good.
 # -*- coding: UTF-8 -*-
 
-from config.base import get_environment_variable, get_python_version, LOGIN_URL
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

PR #1532, opened 2021-01-19
Write new void_api_device_id cookie, it it changes during the sign in process

SailingSteve:
See wevote/WeVoteServer/issues/1508


DaleMcGrew:
👍

PR #1539, opened 2021-02-03
First step in adding 'voter_guide_voterguidepossibility' to fast sync

SailingSteve:
This PR must come first, since the client is not allowed to request any table that the server does not have white listed.


DaleMcGrew:
👍

PR #1565, opened 2021-03-17
Major rework of Subscriptions (Memberships)

SailingSteve:
Intermediate checkin.
The latest "2020" Stripe APIs that we are now using were pretty different from the 2018 API, and allowed for a big simplification, improved reliability, but not much compatibility.  So I abandoned the donate "app" and replaced it with the stripe_donations "app"
Organizational Subscriptions had broken Donation Subscriptions and regular Donations, so Organizational Subscriptions are not implemented in the new stripe_donations app -- not critical path.
As of this checkin, still need:
* a Membership can be created and deleted, but error handling on bad cards is not yet implemented
* Handle login and merging logins of we_vote_ids
* Mobile layout
* Chip in donations

There is a lot of commented out code in stripe_donations, that I will minimize in future PRs.


DaleMcGrew:
👍

PR #1576, opened 2021-03-25
Added option to cleanup Candidate and Politician names on the two template editors

SailingSteve:
Added a new "Change" button to the two editor html renders
A version of the autosize.js library (version from 2013) had been pasted into an editor page with a note of  apology, and had then been cloned 5 times -- removed all of them, and we now load the latest version of autosize from a cdn in template_base.html
Commented out unnecessary error logging, in views_apple.py and in a few other places, to minimize splunk clutter.
Fixed a few other errors that were routinely showing up in splunk.


DaleMcGrew:
👍

PR #1580, opened 2021-04-01
Added a slash to the "/apis/v1/candidateOrPoliticianRepairNames/" API…

SailingSteve:
… call to remove an error that showed up on Dale's local.

Error:
RuntimeError: You called this URL via PUT, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining PUT data. Change your form to point to localhost:8001/apis/v1/candidateOrPoliticianRepairNames/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.


DaleMcGrew:
👍

PR #1600, opened 2021-04-09
New voterSendGoogleContacts API endpoint document page

SailingSteve:
Just a stub, it does not store anything to the DB


DaleMcGrew:
👍

Issue #1611, opened 2021-04-24
Create test for campaignListRetrieve


DaleMcGrew:
You can emulate apis_v1/tests/test_views_candidates_retrieve.py
Please rename to apis_v1/tests/test_views_campaign_list_retrieve.py

This is the expected response documentation: https://api.wevoteusa.org/apis/v1/docs/campaignListRetrieve/
You can see this API being called when you visit: https://campaigns.wevote.us

You should find 13 matches when you search the WeVoteServer code base for `campaignListRetrieve`, and the view can be found in apis_v1/views/views_campaign.py in the function `campaignx_list_retrieve_view`.

Please also see this documentation: https://github.com/wevote/WeVoteServer/blob/develop/docs/README_DJANGO_TESTS.md

PR #1615, opened 2021-04-26
Fixed the TypeError referenced in comment views/views_campaign.py. Added unit test for campaignListRetrieve.

ranadeepmitra21:
While creating a unit test for campaignListRetrieve for Issue #1611, we ran into a type error referenced in a comment on line 30 of "views/views_campaign.py" (see below).  
```
json_string = ''
    try:
        # March 24, 2021: Throwing "TypeError: Object of type 'HttpResponse' is not JSON serializable"
        json_string = json.dumps(json_data)
    except Exception as e:
        status = "Caught error for voter_device_id " + voter_device_id
        handle_exception(e, logger=logger, exception_message=status)

    return HttpResponse(json_string, content_type='application/json')
```

We believe this error was being caused by the API endpoint in "campaign/controllers.py" returning an HttpsResponse rather than the raw JSON data when an invalid voter_device_id was passed into the GET request. This only happened when the branch for an invalid voter_device_id was passed. See the code block below.
```
if not positive_value_exists(voter_results['voter_found']):
        status += 'VOTER_WE_VOTE_ID_COULD_NOT_BE_FETCHED '
        json_data = {
            'status': status,
            'success': False,
            'campaignx_list': [],
        }
        return HttpResponse(json.dumps(json_data), content_type='application/json')
```

We solved this error by replacing  `return HttpResponse(json.dumps(json_data), content_type='application/json')`  with `return json_data`.

After resolving this error, we were able to create a unit test for campaignListRetrieve in "apis_v1/tests/test_views_campaign_list_retrieve.py".

This was worked on by @ranadeepmitra21 and @patkeish

### DIFF
--- a/apis_v1/tests/test_views_campaign_list_retrieve.py
+++ b/apis_v1/tests/test_views_campaign_list_retrieve.py
@@ -1,0 +1,42 @@
+# apis_v1/test_views_campaign_list_retrieve.py
+# Brought to you by We Vote. Be good.
+# -*- coding: UTF-8 -*-
+
+from admin_tools.controllers import import_data_for_tests
+from django.urls import reverse
+from django.test import TestCase
+import json
+from office.models import ContestOffice, ContestOfficeManager
+
+
+class WeVoteAPIsV1TestsCampaignListRetrieve(TestCase):
+    databases = ["default", "readonly"]
+
+    def setUp(self):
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Thank you @ranadeepmitra21 this looks great. Thank you for catching and fixing the bug!

DaleMcGrew:
👍

PR #1618, opened 2021-04-30
Tests for the /positionListForVoter endpoint 

utsab:
**Description of the API Endpoint according to WeVoteServer’s documentation:** 

A list of all positions (support/oppose/info) for this voter relating to Candidates, Measures or Offices.

**How to trigger the API endpoint:** 

Go to wevoute.us/yourpage. 

You can add items to the position_list field by clicking on “Choose” or “Oppose” for a candidate.  

**Test cases that we covered:**

- Request does not contain a valid voter_device_id
- Request contains a valid voter_device_id but there is no registered user associated with the id.
- Suppose the user is registered and clicked on “choose” for one candidate


utsab:
Hi @DaleMcGrew, I tried out a new approach for "stubbing" the database with sample data by directly invoking the model and saving a new record to the table (as opposed to invoking the API endpoint to save new data to the database).  I believe this simplifies the logic of the test (fewer things can go wrong) so that it focuses more precisely on the individual endpoint that is being tested.  I also wanted to do it this way so that my students could get comfortable using the models.  Please let me know if you have any feedback about this approach.

DaleMcGrew:
👍

PR #1634, opened 2021-06-15
Install working on MacOS Big Sur (Probably requires Python 3.9 upgrade on each Mac)

SailingSteve:
New packages are specified in requirements.txt

Added very rough instruction updates in docs/README_MAC_SIMPLIFIED_INSTALL.md
which need to be verified on fresh install on a new mac.


DaleMcGrew:
👍

PR #1635, opened 2021-06-17
Upgraded from Django 2.2.22 to 3.2.4 (Probably requires at least Python 3.6 (possibly 3.8) upgrade on each Mac)

SailingSteve:
The django upgrade looks good.  No manual patching is required.  I can sign in now.


DaleMcGrew:
👍

PR #1636, opened 2021-06-18
Updated the simplified mac install, to match the updated ide, tools, and OS.

SailingSteve:
Ran a test run on a mostly blank Mac.
Cleared the bug where an un-configured stripe secret key, was throwing misleading errors.


DaleMcGrew:
👍

PR #1645, opened 2021-06-23
VoterRetrieve now returns ext_for_map_search to save an API call on first load.  Merged

SailingSteve:
Adds 7ms to the VoterRetrieve call.


DaleMcGrew:
👍

PR #1647, opened 2021-06-26
Fixed the geolocation annotation calculation for Django 3.2

SailingSteve:
Removed the # TODO: Update to a more modern approach? I think this will be deprecated in > Django 1.9


DaleMcGrew:
👍

PR #1649, opened 2021-06-30
New friends_lists_all api

SailingSteve:
Currently all 6 friend requests that are included in this one API fired simultaneously when switching to /news, which tied up all 6 browser http channels for 1.25 seconds.  This new api takes half as long, but more importantly it frees up 5 channels.


DaleMcGrew:
👍

PR #1660, opened 2021-07-09
Support added for donation Chip Ins

SailingSteve:
Requires a migration
Renamed a lot of variables, and some column names, to support Chip In, and to reduce confusion in prep for adding back in premium subscriptions at some point.
This PR does not make any Python template changes for wevote/WeVoteServer/issues/1658
Those changes will follow.  These changes are necessary to support changes coming in the Campaign app side.
WebApp donation changes will also be required.


DaleMcGrew:
👍

PR #1666, opened 2021-07-19
Multiple fixes to Donate.

SailingSteve:
I was reusing the default credit card for every new subscription, this is now fixed.
Consolidated multiple functions that saved subscriptions and payments into two functions that can
handle the pieces of the two records that arrive separately from our API call to the server, and in the various webhooks that arrive after the stripe events that we subscribe to.
Added SQL Transactions (Django transaction.atomic) to the subscription and payment write functions to stop Webhooks that arrive almost at the same instant from colliding and overwriting each other's data when saving to the db.


DaleMcGrew:
👍

PR #1672, opened 2021-07-27
Show the cumulative chip ins on the /campaign/.../supporters template page

SailingSteve:
I spent a number of hours trying to make a filter for chip in donations that worked with the other filters, which requires a join in django, and it didn't seem worth more effort.  If the supporter is displayed, their donation total will display.  I could have done this in SQL, but then it wouldn't have worked with the other filters.
The campaign name is now stored in the chip in donation records (going forward).
Progress towards wevote/WeVoteServer/issues/1658


DaleMcGrew:
👍

PR #1673, opened 2021-07-27
Added a clickable link to cumulative chip ins on the /campaign/.../supporters template page

SailingSteve:
The link turns on the Include Supporters buttons initially, so you will see any supporters with chipins.


DaleMcGrew:
👍

PR #1675, opened 2021-08-16
Get "Refresh Facebook Photo" button working again

SailingSteve:
Rewrote the FB API code to do this, using the latest V11.0 code.
We have some FB API V3 code, that I think has been broken for a while -- didn't remove it.
This change requires two new environment variables, to make V11 server side API queries.


DaleMcGrew:
👍

PR #1678, opened 2021-08-18
Facebook photos refresh improvements

SailingSteve:
Now handles facebook urls with concatenated facebook account names and account ids like https://facebook.com/andrea-nelson-moore-for-judge-145896229538232
The bulk retrieve facebook photos worked as is -- it already used the new code.
Added "Refresh Facebook Photo" to organization_edit.html and organization_position_list.html
Deprecated earlier facebook api env variables, but still allow them to work as undocumented aliases


DaleMcGrew:
👍

PR #1681, opened 2021-08-19
Facebook photos refresh fix

SailingSteve:
Refreshing facebook on the organization_edit page, now has a return url based on the organization sql row id, instead of the organization we_vote_id (and as a result, now returns to the correct page).
Disable the "Refresh Facebook Photo" button on the candidate_edit and the organization_edit page after you change the text of the URL, since otherwise, pressing the button would do a refresh based on the prior value (the "currently in postgres" value) of the url.


DaleMcGrew:
👍

PR #1683, opened 2021-08-20
Facebook social-auth app fix

SailingSteve:
I think this is going to work, the social app requires globals with specific names.


DaleMcGrew:
👍

PR #1688, opened 2021-09-02
Get facebook name, email, and photo into voter if previous attempt failed or was incomplete

SailingSteve:
Previously, if we were not creating a new facebook_link_to_voter, we would not update the voter
with the latest name, email, and photo as a result of a login to facebook.
I suspect that we had not been gathering photos for facebook logins since "10AM PDT on October 23, 2020" when the API started requiring an access token.

Verify that this addresses wevote/Campaigns/issues/42


DaleMcGrew:
👍

PR #1742, opened 2021-11-06
Fixed the new misspelled table names in fast load -- This feature will not work on the local, until this fix gets to the production servers.

SailingSteve:
Also
A bunch of things changed with the simplified install, and it will need another run through on a clean machine to get it right.  Not today...


DaleMcGrew:
👍

PR #1748, opened 2021-11-08
Facebook access tokens are now longer than 256 characters

SailingSteve:
Must be nice to have too many quadrillions of sessions.
Also added a (redundant) "signed_in" field to voterRetrieve to aid in debugging -- all the signin state is in the same area in the response.


DaleMcGrew:
👍

PR #1760, opened 2021-12-15
Upgraded to Twitter API V2, Tweepy V4.4.0

SailingSteve:
The caching of images to AWS really slows things down, even the time to sign in via Twitter.  We should think about putting those into a Queue for asynchronous saving, and decoupling the saving delay (sometimes 5 seconds or more) from the HttpResponse.


DaleMcGrew:
Hi @SailingSteve that is a great idea to decouple photo saving. Saving time on Twitter sign in will help keep voters engaged. 👍

PR #1764, opened 2021-12-17
Fixed the inability to add a new Endorser.

SailingSteve:
Added conditional logging to image/controllers to see where thing are slowing down.
The previous upgrade was actually to Twitter API V1.1 -- V2.0 is not finished enough for us to use.


DaleMcGrew:
👍

PR #1765, opened 2021-12-22
Speed up twitter sign-in (initially only in the WebApp)

SailingSteve:
Broke the twitter signin that used to take 15 seconds or more, the first part twitter_sign_in_retrieve_for_api no longer caches the images, and allows the user to signin in about 3 seconds.  The second part, twitter_process_deferred_images_for_api is fired by the WebApp as soon as the twitter sign in completes, which then processes the images asynchronously.  Then about 15 seconds after the sign in, Ballot fires off a voterRetrieve to replace the anonymous icon with the voter's twitter photo.


DaleMcGrew:
This is such a great upgrade @SailingSteve! 👍

PR #1782, opened 2022-02-07
Allows all customized instances of WebApp and Campaigns to use a conistent index.html

SailingSteve:
site_configuration_retrieve now returns chosen_google_analytics_tracking_id (used to be called chosen_google_analytics_account_number)
organization_index_doc.py now takes a organization_incoming_domain so you can test it with different domains.
Eliminated (mostly duplicate) organization_campaigns_index.html with existing organization_index.html


DaleMcGrew:
👍

PR #1791, opened 2022-03-01
retrieve_twitter_ids_i_follow_from_twitter now uses bearer token

SailingSteve:
Function now iterates for up to 5000 usernames (aka handles or ids) inline without saving the cursor.  Takes between 1/4 and two seconds, but the request is fired off after sign in with twitter, so it does not slow down sign in.
Rewritten in V2 Twitter API


DaleMcGrew:
Hooray! 👍

PR #1800, opened 2022-04-11
Google visualization fixes

SailingSteve:
Handle occasional uninitialized red pin size.
Build correct url in the navigation pop-up for the /e/ href case.


DaleMcGrew:
👍

PR #1802, opened 2022-04-12
Rewrote the geolocation distance calculation to use great circle distances

SailingSteve:
We can now limit to an exact distance, which has been set to 25 miles.
See spec https://docs.google.com/document/d/1xaGiEWQ_FHmgsK2msusGSZcX256kwe4J06IHMyrdf_o/edit#


DaleMcGrew:
👍

PR #1804, opened 2022-04-14
Fix python requirements so deployment works

josephevans:
Update the google package requirements with compatible versions. Code not tested (couldn't find any unit tests to run.)

Also update django, per recommendation (SQL vulnerability.)


DaleMcGrew:
👍

PR #1808, opened 2022-04-25
Update gunicorn

josephevans:
Update gunicorn to latest version.

Update API load tests to work with newer Locust versions.


DaleMcGrew:
👍

PR #1810, opened 2022-04-25
Updated "Facebook Login" from Api 6 to 12 on the server

SailingSteve:
Changes will be needed in your dev setup see ServerOAuth.md
The server now should be run in all cases in ssl (very easy to do) see ServerOAuth.md


DaleMcGrew:
👍

PR #1812, opened 2022-04-27
Updated ServerOAuth.md to cover Twitter. Twitter and Facebook login work locally when the correct environment variables are used.

SailingSteve:
The local server needs to be run in SSL for both types of login, to work.
The production servers can be HTTP, with a load balancer doing the SSL encryption as before.
For twitter SOCIAL_AUTH_TWITTER_KEY **needs to be the same as** "API Key" from the "Consumer Keys" setting on the developer.twitter.com website.
SOCIAL_AUTH_TWITTER_SECRET  **needs to be the same as**  "Secret" from the "Consumer Keys"


DaleMcGrew:
👍

PR #1847, opened 2022-06-16
Sign In With Apple working, Experiment 6 for git compile date

SailingSteve:
There was a (not brilliant) breaking change in PyJWT v2


DaleMcGrew:
👍

PR #1883, opened 2022-07-15
Experiment 25, exploring copy_expert simplification

SailingSteve:
Suspect that there are problems with these tables in the AWS postgres db:
  'position_positionentered'
  'organization_organization'
  'candidate_candidatecampaign'
  'polling_location_pollinglocation'

so running them last, and seeing how far we get


DaleMcGrew:
👍

PR #1887, opened 2022-07-17
Removed the experimental logging in retrieve_tables.  Finished move_voter_guide_possibility_positions_to_requested_voter_guide_possibility

SailingSteve:
Fast load of election data is working again, probably Joe's allowing very long response times on the table load api calls was the fix.


DaleMcGrew:
👍

PR #1890, opened 2022-07-21
Protected against a crash in move_voter_guide_possibility_positions_to_requested_voter_guide_possibility

SailingSteve:
Crash was in a log line, now is in a try/exception block.


DaleMcGrew:
👍

Issue #1888, opened 2022-07-19
Update documentation for "voterEmailAddress..." API calls


DaleMcGrew:
There are four documentation files starting with "voterEmailAddress...":
WeVoteServer/apis_v1/documentation_source/voter_email_address_retrieve_doc.py

Please verify that the documentation is up-to-date with the latest code.

PR #1902, opened 2022-08-02
Email address save doc

Jakaboy3:
#1888
@yuli48


config/base.py
```diff
@@ -11,7 +11,8 @@
 import re
 from django.core.exceptions import ImproperlyConfigured
 from django.db import connection
-
+import sys
```
DaleMcGrew: Thank you Yulisa and Jack! Could you please roll back (undo) your changes to the config/base.py file?

DaleMcGrew (changes requested):
Thank you again Yulisa and Jack, could you also please roll back the changes to package-lock.json?

config/base.py
DaleMcGrew: Could you also please roll back the changes to package-lock.json? Thank you in advance.

Jakaboy3:
@DaleMcGrew We made another pull request with only one file since my partner and I do not know how to roll back the code.

PR #1909, opened 2022-08-19
Adds an app-adds.txt for facebook.

SailingSteve:
Hopefully this will resolve the cordova facebook login problem.


DaleMcGrew:
👍

PR #1915, opened 2022-09-14
Added phone_number and api_type fields for importing contacts from iPhones

SailingSteve:
any iOS device, and Apple Silicon


DaleMcGrew:
👍

PR #1921, opened 2022-10-08
Added the ability to enter alternate chosen_domain_string values (for "www" subdomain or another alternate).

DaleMcGrew:
Added the ability to enter alternate chosen_domain_string values (for "www" subdomain or another alternate).


DaleMcGrew:
👍

PR #1973, opened 2022-12-14
Much faster Sign in with Facebook

SailingSteve:
This code is reverse compatible with the WebApp code that is in upstream/devlop at this moment.  So it can be put into production, even if the wevote.us side lags for days or weeks.
Multiple lint cleanups


DaleMcGrew:
👍

PR #1974, opened 2022-12-21
Process incoming Facebook images in threads, to speed up save/retrieve queries for Facebook data

SailingSteve:
Enhanced delete_all_voter_information_permanently to also delete facebook_users and facebook_link_to_voter so that processing Facebook images in threads can be fully tested for new users. 
Now delete_all_voter_information_permanently logs what voter was deleted, and who deleted them.


DaleMcGrew:
👍

PR #1976, opened 2022-12-23
More logging to see why facebook signin takes so much longer on the live servers

SailingSteve:
Maybe 10x longer.  tbd


DaleMcGrew:
👍

PR #1984, opened 2023-01-02
Added more logging for Stripe webhooks so we can get more data on attacks

SailingSteve:
So we can figure out how to preempt more of them.

Conditionally enabled donationWithStripe, does not 404 if fetch_stripe_processing_enabled_state() is true Added a column to wevote_settings in case we want to add a simple configuration page, in the meantime, added the new configuration to /stripe_donations/suspects_list/

### DIFF
--- a/apis_v1/views/views_donation.py
+++ b/apis_v1/views/views_donation.py
@@ -5,7 +5,7 @@
 import json
 
 import stripe
-from django.http import HttpResponse, HttpResponseNotFound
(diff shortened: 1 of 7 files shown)


SailingSteve:
Screen Shot 2023-01-02 at 3 08 55 PM

DaleMcGrew:
Thank you @SailingSteve!  👍 
I'd like to discuss before pushing live to the API server. I'm working today and am pretty flexible on times to speak.

PR #1991, opened 2023-01-06
Move all SQS stuff into aws directory.

josephevans:
Also define structure for functions to simplify readability.


DaleMcGrew:
👍

PR #2006, opened 2023-01-11
Fix for merge in the main process overwriting with stale data, the faster SQS voter.save().

SailingSteve:
During facebook sign-in with a facebook account with a new avatar image.


DaleMcGrew:
👍

PR #2019, opened 2023-02-01
Limit stripe "card testing" by checking country, and by checking frequency of hits by IP

SailingSteve:
Added a new "app" stripe_ip_history to track loads of the /more/donate page, so we can detect high frequency card testing and then block them
Added a test tool on /stripe_donations/suspects_list to clear all the ip_history records for an ip


DaleMcGrew:
👍

PR #2064, opened 2023-03-17
Potentially working pdf2htmlEX for pdfs served up by Cloudflare (try 2)

SailingSteve:
Now handles 403s from Cloudflare, with a fallback to trying to get the page from the Google search page cache


DaleMcGrew:
👍

PR #2072, opened 2023-03-19
Potentially working pdf2htmlEX for pdfs served up by Cloudflare (try 10)

SailingSteve:
Write temporary pdf file as a string, not binary


DaleMcGrew:
👍

PR #2074, opened 2023-03-19
Fallback to google bypass works for CADEM url, added some minor cleanup

SailingSteve:
It works for now, needs further testing, after which I'll remove the ERROR logging.


DaleMcGrew:
👍

PR #2115, opened 2023-05-20
Second pass at some instrumentation for voterGuidePossibilityHighlightsRetrieve

SailingSteve:
to look at performance issues.
changed info level logging to error so they will show up in CloudWatch


DaleMcGrew:
👍

PR #2126, opened 2023-05-24
Eighth pass at instr. for voterGuidePossibilityHighlightsRetrieve

SailingSteve:
I found my problem.  retrieve_candidates_from_non_unique_identifiers takes two seconds to execute, and it gets called 14 times for a page with 14 possible highlights.   The function does un-indexed searches on 12 text fields over 76,000 records.  I could try indexing all of those fields, and see if it improves.  Indexes take some CPU to maintain, but we probably add candidates pretty rarely.  First I am going to run explains and see more detail.


DaleMcGrew:
👍  Great catch!

PR #2127, opened 2023-05-24
Ninth pass (b) at instr. for voterGuidePossibilityHighlightsRetrieve

SailingSteve:
to look at performance issues.

Added 10 more indexes to CandidateCampaign


DaleMcGrew:
👍  These look like a very good idea. Thank you @SailingSteve.

PR #2139, opened 2023-06-08
deduped voterGuidePossibilityPositionsRetrieve and now excludes records with null ballot_item_names

SailingSteve:
Also added a date_created and date_updated field to VoterGuidePossibilityPosition


DaleMcGrew:
👍

PR #2153, opened 2023-06-29
Find representatives with OCD ID/State mismatch.

jtconnor:
https://wevoteusa.atlassian.net/browse/WV-92

Tested:
Running locally, finding representatives with mismatch, and filtering representatives with mismatch.


DaleMcGrew:
👍  Thank you @jtconnor, I'll check the Representative data with your script, and then we can figure out the next steps on https://wevoteusa.atlassian.net/browse/WV-92
PS. I see all incoming pull requests because I subscribe to our #github Slack channel.

PR #2167, opened 2023-07-12
Find politicians with representatives with OCD ID/State mismatch.

jtconnor:
- Note that we'll need to set `ocd_id_state_mismatch_checked` to False to re-iterate through Representatives and find mismatches and corresponding Politicians.

https://wevoteusa.atlassian.net/browse/WV-92

Tested:
- Running locally and querying all representatives to find mismatches.
- Observing message about politicians found with mismatch.
- Going to Politicians view page and filtering by mismatch.


DaleMcGrew:
👍

PR #2173, opened 2023-07-19
Added gender_guesser for politician records with missing gender

SailingSteve:
The issue asks for the functionality to be added to repair_imported_names, but if you had two different repair criteria each of which could add a name to the list to be worked on, it would be confusing, so I just made another page.
For https://wevoteusa.atlassian.net/browse/WV-87


DaleMcGrew:
👍 Looks great @SailingSteve!

PR #2179, opened 2023-07-26
Adds fips codes, county, and lat/lng coordinates for every voteraddress

SailingSteve:
There are some unincorporated areas and misspelled addresses that it logs an error for. All the other records in the live database could be processed in a half hour or so. 
If we want to move forward with this approach:
* I would duplicate the fields you would like into the voter record.
* Hook it into new address creation, so it fills in fips codes, county, and lat/lng on the fly
* Move the fipsCodesWithPosition.json file into a postgress table for easier maintenance
* Move the unincorporated "cities" (sometimes real estate developments) that are flagged as misses into a table, but they are mixed in with invalid addresses which requires manual work.

https://wevotedeveloper.com:8000/apis/v1/voterUpdateFips/?limit=10000


DaleMcGrew:
Thank you @SailingSteve! Approved. 👍 
Before moving forward on your proposed next steps, let's discuss: https://wevoteusa.atlassian.net/browse/WV-108
and: https://wevoteusa.atlassian.net/browse/WV-113
There are some directions I am taking (in related code) that needs discussion before you make further progress on this area of the code.

PR #2181, opened 2023-07-29
Replaced the match (case) statement in voter/models.py -- match is not available until Python 3.10, and we are on 3.9 in production

SailingSteve:


DaleMcGrew:
👍

PR #2185, opened 2023-07-31
looks like older Python 9, can't handle unused positional data at end of list when unpacking return from function

SailingSteve:


DaleMcGrew:
👍

PR #2189, opened 2023-08-02
One more try to avoid process kill when simply counting the number of record references in a queryset

SailingSteve:
for /apis/v1/voterUpdateFips/?limit=10


DaleMcGrew:
👍

PR #2191, opened 2023-08-04
Improve performance of siteConfigurationRetrieve

SailingSteve:
We had been doing a 3 string un-indexed search even for the wevote.us, and localhost sites, which had been costing us a half of a second on the initial page load.  This was also really distorting the results in Performance and Lighthouse.


DaleMcGrew:
👍

PR #2199, opened 2023-09-02
WV-106: Add map point filters for source_code and WV-107: Add way to show/remove map points based on their source_code

rileychou:
Took code from https://github.com/wevote/WeVoteServer/pull/1925/files, modified it so it uses GOOGLE_MAPS_API_KEY from environment_variables.json.


DaleMcGrew:
👍

PR #2210, opened 2023-10-15
Configured Docker Compose

nyette:
### Changes

Configured Docker Compose

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -21,7 +21,7 @@ services:
     networks:
       - ${COMPOSE_PROJECT_NAME}
     ports:
-      - "5000:5432"
+      - "5432:5432"
     restart: unless-stopped
     user: postgres
     volumes:
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,84 @@
+name: wevote
+
+services:
+  db:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.db
+      tags:
+        - "${COMPOSE_PROJECT_NAME}-db:latest"
+    container_name: "${COMPOSE_PROJECT_NAME}-db"
+    environment:
+      POSTGRES_DB: WeVoteServerDB
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-secret}
+    env_file: .env
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    networks:
+      - ${COMPOSE_PROJECT_NAME}
+    ports:
+      - "5000:5432"
+    restart: unless-stopped
+    user: postgres
+    volumes:
+      - "${COMPOSE_PROJECT_NAME}-postgres-data:/var/lib/postgresql/data"
+    
+  aws:
+    container_name: "${COMPOSE_PROJECT_NAME}-localstack"
+    depends_on:
+      db:
+        condition: service_healthy
+        restart: true
+    environment:
+      DOCKER_HOST: unix:///var/run/docker.sock
+    image: localstack/localstack
+    networks:
+      - ${COMPOSE_PROJECT_NAME}
+    ports:
+      # Gateway
+      - "4566:4566"
+      # External
+      # - "4510-4559:4510-4559"
+    profiles:
+      - optional
+    volumes:
+      - "${COMPOSE_PROJECT_NAME}-localstack:/var/lib/localstack"
+      - /var/run/docker.sock:/var/run/docker.sock
+  
+  api:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.api
+      tags:
+        - "${COMPOSE_PROJECT_NAME}-api:latest"
+    container_name: "${COMPOSE_PROJECT_NAME}-api"
+    depends_on:
+      db:
+        condition: service_healthy
+        restart: true
+    environment:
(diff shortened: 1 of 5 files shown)


docs/README_API_INSTALL_DOCKER.md
```diff
@@ -1,13 +1,11 @@
 # README for Installation with Docker (OSX or Linux)
 [Back to root README](../README.md)
 
-**This installation method only requires Docker to be installed and running. All dependencies (libraries or Python modules) reside within the docker containers. These instructions are intended for advanced users familiar with docker, git and python.""
+Only [Docker Desktop](https://docs.docker.com/get-docker/) is required.
```
josephevans, line 4: For Linux, docker desktop is actually not required, just dockerd.

docs/README_API_INSTALL_DOCKER.md
```diff
@@ -16,34 +14,56 @@ This method requires [Docker](https://docs.docker.com/get-docker/) to be install
...
+
+    # db
+    POSTGRES_PASSWORD="secret"
     ```
 
-3. Start development environment
+3. Configure PostgreSQL in `init.sql`
 
+    ```sql
+    ALTER SYSTEM SET listen_addresses = '*';
```
josephevans, line 30: Sure this is necessary? Using the dev_environment.sh script, they are started in the same docker namespaces and port 5432 (pgsql) is forwarded properly.

compose.yaml
```diff
@@ -0,0 +1,84 @@
...
+    env_file: .env
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    networks:
+      - ${COMPOSE_PROJECT_NAME}
+    ports:
+      - "5000:5432"
```
josephevans: Do we need all these ports?

compose.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - "${COMPOSE_PROJECT_NAME}-postgres-data:/var/lib/postgresql/data"
+    
+  aws:
+    container_name: "${COMPOSE_PROJECT_NAME}-localstack"
+    depends_on:
+      db:
+        condition: service_healthy
+        restart: true
+    environment:
+      DOCKER_HOST: unix:///var/run/docker.sock
```
josephevans, line 37: Can you help me understand what this is for? This sock file is not accessible within the container.

compose.yaml
```diff
@@ -0,0 +1,84 @@
...
+    ports:
+      # Gateway
+      - "4566:4566"
+      # External
+      # - "4510-4559:4510-4559"
+    profiles:
+      - optional
+    volumes:
+      - "${COMPOSE_PROJECT_NAME}-localstack:/var/lib/localstack"
+      - /var/run/docker.sock:/var/run/docker.sock
```
josephevans, line 50: Why do we need to mount the docker socket inside the container?

docker/Dockerfile.db
```diff
@@ -3,3 +3,5 @@ FROM postgres:12.0-alpine
 USER postgres
 
 EXPOSE 5432
+
+COPY init.sql /docker-entrypoint-initdb.d/
```
josephevans, line 7: Isn't the database automatically created by `python manage.py migrate`?

docker/entrypoint_api_server.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env sh
 
+cp ./config/environment_variables-template.json ./config/environment_variables.json
```
josephevans: Perhaps we should check if the file already exists before copying over it? This way any changes locally won't get clobbered.

docs/README_API_INSTALL_DOCKER.md
nyette: That's true. I should have specified that this setup would be platform-agnostic, rather than just for Unix users. Windows users would need to get Docker Desktop.

docs/README_API_INSTALL_DOCKER.md
nyette: According to the [image](https://hub.docker.com/_/postgres) under database configuration: `Important note: you must set listen_addresses = '*' so that other containers will be able to access postgres.`

compose.yaml
nyette: You're right that we just need port 5432.

compose.yaml
nyette: [Mounting the Docker socket `/var/run/docker.sock` as a volume is required for the Lambda service](https://docs.localstack.cloud/getting-started/installation/#starting-localstack-with-docker-compose). [This thread](https://stackoverflow.com/questions/63201603/what-is-the-result-of-mounting-var-run-docker-sock-in-a-docker-in-docker-scen?noredirect=1&lq=1) explains the situation more eloquently than I ever could.

docker/Dockerfile.db
nyette: I have configured it, so that the API will only run when the database is ready. In `init.sql`, we set `listen_addresses = '*'`, so that the API can access it.

compose.yaml
josephevans: Ahh, got it. Thanks for the explanation. I think we're only using localstack for mocking SQS, but in case we use it for more stuff later, this is ready. :)

josephevans:
Thanks for tackling this; it should really help getting developers up and running with minimal effort.

DaleMcGrew:
👍  Thank you @josephevans and @nyette!

PR #2212, opened 2023-10-27
Brandta/wv 92

brandta-1:
created scripts for Candidate and OfficeHeld that record a mismatch between their state codes and their ocd_division_ids (or their politicians') per WV-92


DaleMcGrew:
Thank you @brandta-1! 👍

PR #2215, opened 2023-10-30
Bypass 2FA for SMS when testing by google/apple app store reviewers.

SailingSteve:
The bypass phone number for testing is '+1 808-935-8555' and the default code is 123456.  If "SOCIAL_AUTH_SMS_BYPASS" if defined, the default code can be overridden.


DaleMcGrew:
👍  Thank you @SailingSteve!

PR #2216, opened 2023-10-31
Handled requests from private IP addresses

nyette:
### Changes

Handled requests from [private IP addresses](https://docs.python.org/3.9/library/ipaddress.html#ipaddress.IPv4Address.is_private), due to the [changing IP addresses of Docker containers](https://docs.docker.com/desktop/networking/#changing-internal-ip-addresses)


DaleMcGrew:
👍

compose.yaml
```diff
@@ -64,17 +61,23 @@ services:
...
+      - "127.0.0.1:8000:8000"
     volumes:
       - "./:/${COMPOSE_PROJECT_NAME}"
     working_dir: "/${COMPOSE_PROJECT_NAME}"
 
 networks:
   wevote:
+    ipam:
+      config:
+        - subnet: "172.27.0.0/16"
```
josephevans, line 79: Why not let docker figure out the subnet to use? If this particular subnet is already defined for another container, it will fail to work.

config/base.py
```diff
@@ -140,6 +140,9 @@ def get_postgres_version():
     '127.0.0.1'
 ]
 
+ALLOWED_CIDR_NETS = [
```
josephevans, line 143: I assume this is used in the allow_cidr module. This change would only allow connections from the hard-coded subnet defined above. This won't work for production use.

nyette: I intended to revert these changes, actually. It occurred to me after the fact that we could just use ngrok to rewrite the headers. Thanks for catching this!

PR #2217, opened 2023-10-31
Fix for, if gender guesser has 'unknowns' suppressed via the button you can get 'pages' with no results

SailingSteve:
Fixes newer issue in https://wevoteusa.atlassian.net/jira/software/projects/WV/issues/WV-87
On page:
https://api.wevoteusa.org/politician/set_missing_gender_ids/


DaleMcGrew:
👍

PR #2218, opened 2023-11-02
Updated data deletion flow for voters

nyette:
### Related Issue(s)

[WV-96](https://wevoteusa.atlassian.net/browse/WV-96)

### DIFF
--- a/voter/controllers.py
+++ b/voter/controllers.py
@@ -4,6 +4,7 @@
 import base64
 import json
 import re
+from background_task import background
 from datetime import timedelta
 from io import BytesIO
 from time import time
@@ -141,8 +142,10 @@ def add_state_code_for_display_to_voter_list(voter_we_vote_id_list=None):
     }
     return results
 
-
-def delete_all_voter_information_permanently(voter_to_delete=None, user=None):  # voterDeleteAccount
+# voterDeleteAccount
+# Delete data 5 minutes later
+@background(schedule=300)
+def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
     success = True
     status = ""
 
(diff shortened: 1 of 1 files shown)


voter/controllers.py
```diff
@@ -141,8 +142,10 @@ def add_state_code_for_display_to_voter_list(voter_we_vote_id_list=None):
     }
     return results
 
-
-def delete_all_voter_information_permanently(voter_to_delete=None, user=None):  # voterDeleteAccount
+# voterDeleteAccount
```
DaleMcGrew, line 145: Please move the API name back to be on the same line as the snake-case function name. I maintain this standard so the function name is visible in searches for the API name.

voter/controllers.py
```diff
@@ -141,8 +142,10 @@ def add_state_code_for_display_to_voter_list(voter_we_vote_id_list=None):
...
+# Delete data 5 minutes later
+@background(schedule=300)
```
DaleMcGrew, line 147: Please remove the background delay. I don't want to introduce an individual-server-managed-delay that has no backup if the server were to crash within those 5 minutes.

voter/controllers.py
```diff
@@ -332,70 +335,37 @@ def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
...
-            'email_owner_voter_found': email_owner_voter_found,
-            'facebook_owner_voter_found': facebook_owner_voter_found,
-            'invitation_owner_voter_found': False,
-        }
-        return error_results
-
-    # We need this below
-    voter_device_link = voter_device_link_results['voter_device_link']
-
+# voterDeleteAccount
```
DaleMcGrew, line 338: Please move the API name back to be on the same line as the snake-case function name. I maintain this standard so the function name is visible in searches for the API name.

voter/controllers.py
```diff
@@ -332,70 +335,37 @@ def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
...
-def voter_delete_account_for_api(  # voterDeleteAccount
-        voter_device_id=''):
-    current_voter_found = False
-    email_owner_voter_found = False
-    facebook_owner_voter_found = False
-    twitter_owner_voter_found = False
-    invitation_owner_voter_found = False
-    new_owner_voter = None
-    success = False
-    status = ""
```
DaleMcGrew, line 345: I would like to maintain the pattern of declaring a local status at the top of the file so we can update the status string throughout the function, and pass the accumulated comments out when we exit.

voter/controllers.py
```diff
@@ -332,70 +335,37 @@ def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
...
-        }
-        return error_results
-
-    # We need this below
-    voter_device_link = voter_device_link_results['voter_device_link']
-
+# voterDeleteAccount
+def voter_delete_account_for_api(voter_device_id=''):
+    error_results = {
+        'success': False,
```
DaleMcGrew, line 341: As a design pattern, please include a "status" field even if empty at this point. The reason is that we can get crashing errors if status is missing when the calling view/controller/function expects a status, and it isn't there. I know we add it below, but over the lifecycle of the code, this function may evolve, and another engineer might try to pass "error_results" out without adding a status field.

voter/controllers.py
```diff
@@ -332,70 +335,37 @@ def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
...
-    results = delete_all_voter_information_permanently(voter_to_delete=voter)
+    # If none of the keys are valid, error_results['status'] = "VOTER_SECRET_KEY_NOT_PASSED_IN"
+    # return error_results
+    
+    voter_to_delete = voter_results['voter']
 
+    delete_all_voter_information_permanently(voter_to_delete)
+    
+    results = {
+        'status': "DELETE_VOTER ",
```
DaleMcGrew, line 364: I would like to stick with the design pattern of the status message being added to inline, and then passed out as:
'status': status.
In this case it may feel like overkill, but it makes extending the code later much faster during code maintenance.

voter/controllers.py
```diff
@@ -332,70 +335,37 @@ def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
...
-    #     }
-    #     return error_results
-
-    results = delete_all_voter_information_permanently(voter_to_delete=voter)
+    # If none of the keys are valid, error_results['status'] = "VOTER_SECRET_KEY_NOT_PASSED_IN"
+    # return error_results
+    
+    voter_to_delete = voter_results['voter']
 
+    delete_all_voter_information_permanently(voter_to_delete)
```
DaleMcGrew, line 361: Please capture the status text coming from delete_all_voter_information_permanently, add it to the status value built up within this function, so it can be passed out of this function. 

In general, be cautious about changing code where we are passing results from one function out of the wrapping function. For example, you are also potentially masking a success value of False and replacing it with a blanket "True". I'd like to maintain passing success = False out of voter_delete_account_for_api through if success = False coming out of delete_all_voter_information_permanently.

voter/controllers.py
```diff
@@ -332,70 +335,37 @@ def delete_all_voter_information_permanently(voter_to_delete=None, user=None):
...
-            'current_voter_found': current_voter_found,
-            'email_owner_voter_found': email_owner_voter_found,
-            'facebook_owner_voter_found': facebook_owner_voter_found,
-            'invitation_owner_voter_found': False,
-        }
+        error_results['status'] = "VOTER_NOT_FOUND_BY_VOTER_ID"
         return error_results
 
-    voter = voter_results['voter']
-    status += "DELETE_VOTER-" + str(voter.we_vote_id) + " "
```
DaleMcGrew, line 380: Please maintain this status addition and include the voter_we_vote_id since it could be very helpful during debugging:
status += "DELETE_VOTER-" + str(voter.we_vote_id) + " "

DaleMcGrew (changes requested):
Thank you @nyette for these updates. It is great to have you working across both code bases.

PR #2255, opened 2023-11-28
Fixes for starting up Google Vertex AI

SailingSteve:
This PR can work with the oauth secret file that is used for Firebase, but the project code number is different, so all the
Vertex config (from the Google console) would have to be moved to the Firebase project.  So it probably is easier to just have two oauth files on the production servers (the pre-existing Firebase file "/etc/wevoteapps-f4b59fec8c57.json" and a new one for Vertex AI.)

After misc fixes, the error response when using the Firebase OAuth file is ...
```
Error response from Vertex Model: 403 Permission 'aiplatform.endpoints.predict' denied on resource '//aiplatform.googleapis.com/projects/we-vote-ballot/locations/us-west1/publishers/google/models/text-bison@latest' (or it may not exist). [reason: "IAM_PERMISSION_DENIED"
domain: "aiplatform.googleapis.com"
metadata {
  key: "resource"
  value: "projects/we-vote-ballot/locations/us-west1/publishers/google/models/text-bison@latest"
}
metadata {
  key: "permission"
  value: "aiplatform.endpoints.predict"
}
]
```

When I get home tomorrow afternoon, I'll work on generating a new Oauth file for the Vertex project for the "We Vote Ballot" project.


DaleMcGrew:
👍

PR #2258, opened 2023-11-29
Google Vertex AI (with new secret JWT) can now make 'predict' requests

SailingSteve:
Auth is working.
It doesn't find names, but it looks like the predict query isn't using the scrape data.
Feel free to remove any logging or print statements that I added.


DaleMcGrew:
👍 Thank you @SailingSteve for getting us over this hurdle!

PR #2262, opened 2023-12-04
WV-103_add Friends Only checkbox

stilwellmg:
In the file templates/position/position_list.html:
1. Add a checkbox labeled "Friends Only" that is visible only when the user role='admin'

In the file position/views_admin.py:
1. Add an attribute named "show_admin_options" and default to False. Change the value to True if the user is assigned the role 'admin'
2. Change page behavior so only Public opinions are displayed unless the Friends Only box is checked.
3. Change the page behavior to only display Friends opinions when the users role='admin' and the Friends Only box is checked.
4. Set maximum rows of output to 20.
5. Add the same conditions for displaying Public or Friends Only statistics when the Show Statistics box is checked.


DaleMcGrew:
Thank you @stilwellmg! 👍

PR #2271, opened 2023-12-11
Brandta/wv-118

brandta-1:
initial commit for Issue WV-118, the algorithm used is simply the resize method from PILLOW using nearest neighbor resampling, which generates a color that represents the dominant color in the sample (the top left 20x20 pixel box in a large profile image)

This is quite slow, at about 1 second per politician, so it would be worth exploring other methods.


DaleMcGrew:
👍  Thank you Alex, I look forward to discussing.

PR #2278, opened 2023-12-21
Edit Voter page: Can now generate password quickly

SailingSteve:
Some cleanup of the screen layout
Pycharm was complaining about some broken css, that looked low value, so I commented it out.

Resolves https://wevoteusa.atlassian.net/browse/WV-188


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #2283, opened 2023-12-24
Added google verification file to Python server, so that it can serve google site files for the Webapp

SailingSteve:


DaleMcGrew:
👍

PR #2287, opened 2024-01-09
Added Has Email checkbox on Candidates Admin page (WV-93)

hesamshayegan:


candidate/views_admin.py
```diff
@@ -482,6 +484,22 @@ def candidate_list_view(request):
             populate_contest_office_data_status += \
                 "SCRIPT: {entries_to_process:,} entries to process (populate_contest_office_data). " \
                 "".format(entries_to_process=candidate_ultimate_count) + " "
+
+        # Filter a queryset of candidtes based on whether they have email-related fields with a length greater than 2
+        if positive_value_exists(show_candidates_with_email):
+            candidate_query = \
+                candidate_query.annotate(candidate_email_address_length=Length('candidate_email_address'))
+            candidate_query = candidate_query.annotate(candidate_email_length=Length('candidate_email'))
+            candidate_query = candidate_query.annotate(candidate_email2_length=Length('candidate_email2'))
```
DaleMcGrew: Hello @hesamshayegan, for the Candidate object in WeVoteServer/candidate/models.py we only have one field (no "candidate_email2" or "candidate_email3".

templates/candidate/candidate_list.html
```diff
@@ -328,6 +328,10 @@ <h1>Candidates</h1>
     <br />
 
     {# Turn on/off review mode #}
```
DaleMcGrew: Please move this "Turn on/off review mode" comment to be above "review_mode_id" code, and add a new comment here saying something like "Limit list to only show candidates with an email address".

DaleMcGrew (changes requested):
Thank you @hesamshayegan, this is almost there, pending my two comments. Great work!

hesamshayegan:
Hi @DaleMcGrew, thanks for your comments. I updated candidate_list.html in templates and removed the second and third emails from views_admin.py.

DaleMcGrew:
Thank you @hesamshayegan! 👍

PR #2295, opened 2024-01-25
Need to get ip from HTTP_X_FORWARDED_FOR in production -- googlebot_site_map

SailingSteve:
Since the Python server is behind a webserver


DaleMcGrew:
👍

PR #2303, opened 2024-01-31
Brandta/wv 189

brandta-1:
updated color generation controller, updated admin edit page for politician, and admin edit process view controller


DaleMcGrew:
👍

PR #2305, opened 2024-02-02
updating server set up instructions

lredmo1:
I ran into a lot of issues trying to get set up on Linux. I did it spread out over some time and unfortunately was not taking notes early on about the issues I was running into. So these changes primarily reflect some of the workarounds I had to do in the later steps. I did run into some issues early on, but don't have good notes on how I fixed them, unfortunately. 

These are all just suggestions, so please disregard any that seem incorrect or confusing.

### DIFF
--- a/docs/README_API_INSTALL_POSTGRES_LINUX.md
+++ b/docs/README_API_INSTALL_POSTGRES_LINUX.md
@@ -4,17 +4,19 @@
 
 ## Installing PostgreSQL on Linux
 
-    $ sudo apt-get install python-psycopg2 
-    $ sudo apt-get install python-dev
-    $ sudo apt-get install postgresql-client
-    $ sudo apt-get install postgresql postgresql-contrib
-    $ sudo -u postgres psql postgres
-    postgres=# \password postgres
-    (NOTE: We recommend using a very simple password like "123" since it is just local development)
-    postgres=# \q
-    $ sudo -u postgres createdb WeVoteServerDB
+(https://www.postgresql.org/download/linux/ubuntu/)
+	# Create the file repository configuration:
+	$ sudo sh -c 'echo "deb https://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
 
-See here for more detailed [instructions](https://help.ubuntu.com/community/PostgreSQL)
+	# Import the repository signing key:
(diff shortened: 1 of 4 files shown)


docs/README_API_INSTALL_POSTGRES_LINUX.md
```diff
@@ -4,17 +4,18 @@
 
 ## Installing PostgreSQL on Linux
 
-    $ sudo apt-get install python-psycopg2
```
lredmo1, line 7: This script generated a lot of errors for me. Unfortunately I don't remember how I solved it because I think I did it on a call with Nette. I was able to piece together a few things from Slack related to that convo, like links she sent me. The new instructions here are taken from one of the [links](https://www.postgresql.org/download/linux/ubuntu/) she sent me.

I *think* these instructions are all I did, but again am not 100% sure. So please pay special attention to this section and edit as needed.

lredmo1: In case the context is helpful, this was my slack convo with Nette about this, before she sent me the above link:

> When I run this I get an error `E: Unable to locate package python-psycopg2.`
> 
> I've tried reading up on stack overflow and looking at some of the documentation but without success. Here's what I've tried:
> 
> https://www.psycopg.org/docs/install.html 
> Tried installing binary, which I can do successfully, but then the second step (`$ sudo apt-get install python-dev`) still fails
> 
> https://zoomadmin.com/HowToInstall/UbuntuPackage/python3-psycopg2 
> Tried installing python3-psycopg2, but second step still fails
> 
> https://stackoverflow.com/questions/14391804/e-unable-to-find-a-source-package-for-psycopg2-or-use-pip-with-virtualenv Tried updating `/etc/apt/sources.list` but no change

docs/README_API_INSTALL_PYTHON_LINUX.md
```diff
@@ -42,7 +42,7 @@ Now activate this new virtual environment for WeVoteServer:
 
 ## Installing Python 3 on Linux
 
-    (WeVoteServer) $ pip3 install django-toolbelt
```
lredmo1, line 45: I think this package my be outdated. This script always gave me an error and I was able to get everything installed by just installing django

DaleMcGrew:
Thank you @lredmo1 I appreciate you taking the time to improve the documentation for future engineers! 👍

PR #2322, opened 2024-02-19
googlebot_site_map now only accessed via webapp server

SailingSteve:
so urls in sitemap_index.xml have been updated with weapp base urls.


DaleMcGrew:
👍

PR #2328, opened 2024-02-24
WV-210 extract_twitter_handle_from_text_string function preserves case of Twitter Handle TEAM REVIEW

tejasaditya:
extract_twitter_handle_from_text_string function is updated to do a case-insensitive match and strip the string, preserving the original case of the final remaining string which is the Twitter handle.


DaleMcGrew:
👍 Congratulations on your first pull request Tejas!

PR #2339, opened 2024-02-28
Brandta/WV-274

brandta-1:
# Changes
```politician_list_view``` in ```politician/views_admin.py``` now updates background colors for a given state of politicians.

## Select state from admin view dropdown
state

## Two booleans ```generate_backgrounds``` and ```updates_needed``` must be true

booleans

## Update politicians that need their pictures changed

running


DaleMcGrew:
Thank you @brandta-1! I will play with it and then turn on your code so it can run the changes on the live server. 👍

PR #2343, opened 2024-02-29
bugfix/wv-239 TEAM REVIEW

pbnjcub:
refactored all methods that could be static to static.


ballot/models.py
```diff
@@ -251,7 +257,9 @@ def refresh_cached_ballot_item_office_info(self, ballot_item, contest_office=Non
 
         return ballot_item
 
-    def retrieve_ballot_item(self, ballot_item_id=0):
+    @staticmethod
+    def retrieve_ballot_item(
```
DaleMcGrew: Hi @pbnjcub this works, but we only need to break variables onto a second line when there are enough incoming variables to cause a line wrap.

ballot/models.py
```diff
@@ -702,7 +722,8 @@ def delete_ballot_item(self, ballot_item_id=0):
         }
         return results
 
-    def refresh_all_ballot_item_measure_entries(self, contest_measure):
+    def refresh_all_ballot_item_measure_entries(
+        contest_measure):
```
DaleMcGrew: Missing @staticmethod

ballot/models.py
```diff
@@ -668,7 +686,9 @@ def create_ballot_item_row_entry(
             }
         return results
 
-    def delete_ballot_item(self, ballot_item_id=0):
+    @staticmethod
+    def delete_ballot_item(
```
DaleMcGrew: We only need to break variables onto a second line when there are enough incoming variables to cause a line wrap.

ballot/models.py
```diff
@@ -733,7 +754,9 @@ def refresh_all_ballot_item_measure_entries(self, contest_measure):
             }
         return results
 
-    def refresh_all_ballot_item_office_entries(self, contest_office):
+    @staticmethod
+    def refresh_all_ballot_item_office_entries(
```
DaleMcGrew: We only need to break variables onto a second line when there are enough incoming variables to cause a line wrap.

ballot/models.py
```diff
@@ -966,7 +991,10 @@ def delete_all_ballot_items_for_voter(self, voter_id, google_civic_election_id):
         }
         return results
 
-    def retrieve_ballot_items_for_election(self, google_civic_election_id, state_code=''):
+    @staticmethod
+    def retrieve_ballot_items_for_election(
```
DaleMcGrew: We only need to break variables onto a second line when there are enough incoming variables to cause a line wrap.

DaleMcGrew:
Hi @pbnjcub, thank you for your changes. There was one missing @staticmethod, but ~15 instances where you should leave the incoming variables on one line instead of wrapping to additional lines. Could you please adjust? Please let me know if you have any questions. (I would recommend re-reading the issue description for clarity now that you have been through the code.)

ballot/models.py
pbnjcub: Thanks Dale, I think I updated all of your corrections.

DaleMcGrew:
Great work @pbnjcub! 👍

PR #2344, opened 2024-03-01
WV-284 Admin login page email field should not be case sensitive MERGE READY

smcvey7:
Django authenticate cannot be used with iexact, so I edited the login to first find the user with email__iexact  and save it to user_obj. Then I pass the actual email found in the user_obj to authenticate()

Also fixes 2 issues "PEP 8 E501 line too long"

### DIFF
--- a/admin_tools/views.py
+++ b/admin_tools/views.py
@@ -1741,8 +1741,16 @@ def login_we_vote(request):
             voter_on_stage = request.user
             voter_on_stage_id = voter_on_stage.id
     elif request.POST:
-        username = request.POST.get('username').lower()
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
@smcvey7 have you confirmed this fixes the conditions described in the issue? You can test by creating a new voter account using the "Create New Voter Account" on the /voter/ page on your local Admin server, and then trying to sign in with another browser. Example email address: joe.schmo@WeVoteEducation.org

smcvey7:
I did, and it appears to work correctly.

DaleMcGrew:
👍

PR #2347, opened 2024-03-04
Brandta/WV-274

brandta-1:
## ```politician/controllers_generate_color.py```
Performance bottleneck identified, URL request method changed to be slightly faster:
### Before

requests_test

### After

urllib_test

## ```politician/views_admin.py```

```update_politicians_profile_image_background_color_view``` now checks state code and updates 1000 at a time

## ```templates/politician/politician_list.html```

updated ```a``` tag to match controller

## ```templates/politician/politician_edit.html```

form input for background color modification no longer renders if the politician does not have a picture


DaleMcGrew:
Thank you @brandta-1. 👍

PR #2350, opened 2024-03-04
twitter auth Experiment 2b

SailingSteve:
See what is in the WSGIRequest on the server, to develop new redirect url


DaleMcGrew:
👍

PR #2351, opened 2024-03-04
twitter auth Experiment 3

SailingSteve:
error log lines are not working, but print does ... to cloudwatch See what is in the WSGIRequest on the server, to develop new redirect url


DaleMcGrew:
👍

PR #2354, opened 2024-03-05
twitter auth Experiment 6

SailingSteve:
We are getting "TWITTER_RATE_LIMIT_ERROR: 429 Too Many Requests" errors in production.  And none of my newer log lines except "request.path" and being logged, which indicates some sort of bug, or misconfiguration. 

It still is working perfectly on my local API server and WebApp server with developer Twitter credentials.

Added config logging to confirm that we are configured correctly.

### DIFF
--- a/wevote_social/middleware.py
+++ b/wevote_social/middleware.py
@@ -4,16 +4,14 @@
 
 """Social middleware"""
 
-from django.http import HttpResponse
-from django.shortcuts import redirect
-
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Hi @SailingSteve we may have maxed out our Twitter API calls for the last 24 hours. I'm not sure what we can do about this in the long run, but for now you can probably shift your focus to another issue unless you have other ideas. I'll post this pull request now. 👍

PR #2357, opened 2024-03-06
[WV-237] Convert methods with warning to @staticmethod TEAM REVIEW

smcvey7:
Found warnings in WeVoteServer/api_internal_cache/models.py and converted indicated methods to staticmethods, removing self parameter and moving parameter lists to one line when possible.


DaleMcGrew:
👍  Great work!

PR #2358, opened 2024-03-06
WV-240 Convert methods with warning to @staticmethod

smcvey7:
Found warnings in WeVoteServer/candidate/models.py and converted indicated methods to staticmethods, removing self parameter.

In line 2120 added read_only param to docstring

Fixed PEP 8 line too long warnings


DaleMcGrew:
👍

PR #2360, opened 2024-03-06
[WV-243] Convert methods with warning to @staticmethod TEAM REVIEW

smcvey7:
Found warnings in WeVoteServer/wevote_settings/models.py and converted indicated methods to staticmethods, removing self parameter

### DIFF
--- a/wevote_settings/models.py
+++ b/wevote_settings/models.py
@@ -635,8 +635,11 @@ class RemoteRequestHistoryManager(models.Manager):
         return results
 
     @staticmethod
-    def remote_request_history_entry_exists(kind_of_action, google_civic_election_id,
-                                            candidate_campaign_we_vote_id='', organization_we_vote_id=''):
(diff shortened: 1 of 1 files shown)


wevote_settings/models.py
```diff
@@ -630,7 +634,8 @@ def create_remote_request_history_entry(
         }
         return results
 
-    def remote_request_history_entry_exists(self, kind_of_action, google_civic_election_id,
+    @staticmethod
+    def remote_request_history_entry_exists(kind_of_action, google_civic_election_id,
```
DaleMcGrew: Since you are moving some of these variables to the next line, please put each variable on its own line with 8 characters indented.

smcvey7: Oh gosh, I missed that one! Thanks!

DaleMcGrew:
👍  Thank you @smcvey7!

PR #2361, opened 2024-03-07
 TEAM REVIEW WV-293 updated static methods in WeVoteServer/reaction/models.py fixed static methods

lredmo1:
Updated static methods and styling for parameters that cause text to wrap in WeVoteServer/reaction/models.py.

Updated 3/6: Changed tabs to spaces. Removed self param.


DaleMcGrew:
👍

PR #2364, opened 2024-03-08
[WV-245] Convert methods with warning to @staticmethod TEAM REVIEW

smcvey7:
Removes self parameter for functions with warning and converts to @staticmethod.

Also fixes
-PEP 8 E713 "Test for membership should be 'not in'"
-Missing param in docstring

### DIFF
--- a/voter/models.py
+++ b/voter/models.py
@@ -148,7 +148,8 @@ class VoterManager(BaseUserManager):
         }
         return results
 
-    def alter_linked_organization_we_vote_id(self, voter, linked_organization_we_vote_id=None):
+    @staticmethod
+    def alter_linked_organization_we_vote_id(voter, linked_organization_we_vote_id=None):
         status = ""
         success = True
         if voter and hasattr(voter, "first_name"):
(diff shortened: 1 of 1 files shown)


voter/models.py
```diff
@@ -1622,20 +1640,31 @@ def retrieve_voter_by_twitter_id_old(self, twitter_id):
...
         voter_id = ''
         voter_manager = VoterManager()
         return voter_manager.retrieve_voter(voter_id, primary_email_we_vote_id=primary_email_we_vote_id,
                                             read_only=read_only)
 
-    def retrieve_voter(self, voter_id, email='', voter_we_vote_id='', twitter_request_token='', facebook_id=0,
-                       twitter_id=0, sms='', organization_we_vote_id='', primary_email_we_vote_id='', read_only=False):
+    @staticmethod
+    def retrieve_voter(
+            voter_id, email='',
```
DaleMcGrew, line 1659: Please put each variable on its own line.

DaleMcGrew (approved):
Thank you @smcvey7 I'm approving this pull request, but just had one minor comment for when you check in code later. Thank you! 👍

voter/models.py
smcvey7: Got it, thanks!

PR #2366, opened 2024-03-08
More data stored to twitterapi counter to better understand the Twitter rate limit problem

SailingSteve:
There are a number of temporary print() everywhere we use the API to do OAuth, in case we are getting charged for those too.

The rate limit problem has to be solved before I can make much headway on Sign In With Twitter in production.
Twitter sign-in worked on my local a week ago, and now it doesn't -- I'll keep working on that.


DaleMcGrew:
👍

PR #2375, opened 2024-03-11
WV246 WeVoteServer/Donate/models.py: Convert methods with warning to @staticmethod

fdiazq:
Placed parameters on separate lines for WeVoteServer/Donate/models.py


DaleMcGrew:
👍

PR #2378, opened 2024-03-12
Show the twitter api usage stats displays date in local timezone

SailingSteve:
This api call relies on the TWITTER_BEARER_TOKEN being configured for the correct account.
Maybe the stats are only based on BEARER_TOKEN requests, which we have none.  I'll try the other OAuth 1.0a tomorrow.


DaleMcGrew:
👍

PR #2380, opened 2024-03-12
Alternate auth for the twitter api usage stats

SailingSteve:
We are currently showing zero twitter api usage, which is not correct
There is a line on https://developer.twitter.com/en/docs/twitter-api/v1/rate-limits That says about bearer token auth for the api call... "This limit is considered completely separate from per-application Bearer Token limits." 

So switched to auth based on TWITTER_CONSUMER_KEY and TWITTER_CONSUMER_SECRET as an experiment.

ALSO...
I noticed that redirect_state is now (mysteriously) included again in the redirect from twitter to 
https://api.wevoteusa.org/complete/twitter/?redirect_state=tXhUSmxXzuD8Z5ljtrjcT1bN7aBPjT7X&oauth_token=Ta-OGgAAAAABZ4LOAAABjjXbDZo&oauth_verifier=iqdtM4X8VqX69rz8XC0bh4FBSHYLMukL
So took out my code that attempted to deal with the lack of a redirect_state by returning a simulated response in middleware.com (and added some more logging), I suspect/hope that this will allow the sign in with twitter to work on the server.

### DIFF
--- a/twitter/functions.py
+++ b/twitter/functions.py
@@ -95,7 +95,10 @@ def expand_twitter_public_metrics(twitter_dict):
 
 def retrieve_twitter_rate_limit_info():
     try:
-        auth = tweepy.OAuth2BearerHandler(TWITTER_BEARER_TOKEN)  # TODO this might need to be NOT bearer token based https://developer.twitter.com/en/docs/twitter-api/v1/rate-limits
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #2381, opened 2024-03-13
[WV-235] Convert methods with warning to @staticmethod TEAM REVIEW

jenslawless:
Found warnings in WeVoteServer/campaign/models.py and converted indicated methods to staticmethods, removing self parameter


DaleMcGrew:
👍 Thanks @jenslawless!

PR #2386, opened 2024-03-15
[WV-236] Convert static methods TEAM REVIEW

smcvey7:
Converts indicated methods to static methods and removes self parameter

Also corrects some PEP 8 indentation issues


DaleMcGrew:
👍  Thank you @smcvey7!

PR #2390, opened 2024-03-20
MERGE READY wv-317 Update candidate edit page to display form inputs on the same line

lredmo1:
Update 3/22/24: Moved `Facebook URL` and `Instagram Handle` form groups to their own line, moved `YouTube URL` to its own line, and fixed alignment of `Candidate Has Withdrawn From Election` so that all labels and inputs are vertically centered/aligned with each other.

image
image


DaleMcGrew:
👍

PR #2393, opened 2024-03-21
[WV-248] Convert static methods TEAM REVIEW

smcvey7:
Convert indicated methods to static methods removing self parameter and reorganizing other parameters as needed.


DaleMcGrew:
👍

PR #2396, opened 2024-03-21
Update dependencies for Python 3.12.2 (Part 2)

SailingSteve:
Some more depencies that you only run into when running `python manage.py makemigrations`

See https://wevoteusa.atlassian.net/jira/software/projects/WV/issues/WV-327

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 boto3==1.34.66
 CacheControl==0.13.1
 cloudscraper==1.2.68
-cryptography==42.0.5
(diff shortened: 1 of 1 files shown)


SailingSteve:
The Snyk test fails with a cryptic "requirements.txt  ERROR Failed to process file"
After many experiments I eventually tried falling back to he 3/4/24 version then the 2/19/24 versions and all failed in the same way.
So even though the timing is suspicious, it seems like there is a snyk bug.

DaleMcGrew:
👍

PR #2403, opened 2024-03-25
Fast Load Data loading should state how long it really takes (90% done)

SailingSteve:
Need to update the Master so I can work against real data to fine tune the estimated completion algorithm 

ballot_ballotitem has 17 million records, and is breaking fast load -- Might need to break load request into pieces so they can be redone if the ssl connection times out.

See https://wevoteusa.atlassian.net/issues/WV-207


DaleMcGrew:
👍

PR #2414, opened 2024-03-30
[WV-334] Convert Form Layout Changes to Include Templates - TEAM REVIEW

lredmo1:

Questions:
- district name dict value
- contest office name dict value
- vote usa office dict value
- vote usa politician dict value
- how to do labels with links and extra logic


DaleMcGrew:
👍  This is some great work @lredmo1.

PR #2425, opened 2024-04-02
[WV-294] staticmethods formatting requests

alc0ve:
Edited to format requests and then some! I fixed some spacing as well.

### DIFF
--- a/activity/controllers.py
+++ b/activity/controllers.py
@@ -23,8 +23,7 @@ from voter.models import \
     NOTIFICATION_VOTER_DAILY_SUMMARY_EMAIL, NOTIFICATION_VOTER_DAILY_SUMMARY_SMS, \
     VoterDeviceLinkManager, VoterManager
 import wevote_functions.admin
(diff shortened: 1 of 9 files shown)


representative/models.py
```diff
@@ -508,10 +507,10 @@ def retrieve_polling_location_we_vote_id_list_from_representatives_are_missing(
         # status += "PL_LIST: " + str(polling_location_we_vote_id_list) + " "
         polling_location_we_vote_id_list_found = positive_value_exists(len(polling_location_we_vote_id_list))
         results = {
-            'success':                                  success,
```
DaleMcGrew: Hi @alc0ve we don't want to remove this spacing. The reason we line up the values vertically, instead of removing the spaces like you have done, is for readability, and so that if there are code changes through bulk changes, we can use the introduced misalignment to double-check bugs weren't introduced accidentally. Could you please roll back these specific spacing changes. The work you did to wrap incoming variables (like on line 725 for example) looks good.

alc0ve: I will change it back!!! Thanks!

PR #2426, opened 2024-04-02
REQUIRES PYTHON 3.11 -- Upgrade Django to V5, cryptograph to 42, and Pillow to 10 (2)

SailingSteve:
Requires Python 3.11 (I am using the latest from that branch, 3.11.8) 

Strongly recommend using pyenv/pyenv to change versions, it is much better than using homebrew to upgrade. 

Disabled the not used 'scheduled_tasks'

More improvements in fast_load from master servers, probably 95% done.  It works, but there is a display of "time to done" glitch.  Need to test when the "Master" side is deployed in production.

Snyk recommends setuptools [69.2.0](https://security.snyk.io/package/pip/setuptools/69.2.0), but fails with errors.  So backed down to 65.5.1 which is the minimum they will accept.  Seems like a Snyk bug.


nyette:
In addition to those changes:

`psycopg2-binary>=2.9.5`

[2.9.5 and beyond add support for Python 3.11](https://www.psycopg.org/docs/news.html#what-s-new-in-psycopg-2-9-5)

DaleMcGrew:
Thank you @SailingSteve for this massive upgrade to Python 3.11 and the Django upgrade. 👍

PR #2427, opened 2024-04-03
[WV-256] Convert static methods TEAM REVIEW

smcvey7:
## Fixes

- Converts static methods removing self parameter.
- Tries out option fixing "default value mutable" warnings by initially setting default value to None and then setting the value to [] within the method if the value is None.

## Also addresses

- PEP8 E251 unexpected spaces around keyword/parameter equals
- PEP8 W391 blank line at end of file

## Questions

- This file had several warnings for "access protected member _json of a class". I'm not familiar with the Twitter API, but I was wondering if a potential fix might be replacing:

   ` if "retweeted_status" in tweet_json._json:`
   with:
   `if tweet_json.retweeted_status:`

- I have also seen in this file and others a warning for "variable referenced before assignment". I believe this usually happens when a variable is declared in a try/except clause and then is referenced later in the method. I didn't want to mess around with them (declaring them earlier) without knowing what default value they should have, because it seems like this could introduce bugs.

### DIFF
--- a/twitter/models.py
+++ b/twitter/models.py
@@ -168,7 +168,8 @@ class TwitterUserManager(models.Manager):
     def __unicode__(self):
         return "TwitterUserManager"
 
-    def update_or_create_tweet(self, tweet_json, organization_we_vote_id):
+    @staticmethod
+    def update_or_create_tweet(tweet_json, organization_we_vote_id):
         """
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍

PR #2430, opened 2024-04-03
added (WV-340) + (WV-341)

hesamshayegan:
…delete + (WV-340) Update delete_friends_for_voter to use bulk delete


DaleMcGrew:
👍 Thank you Hesam!

PR #2445, opened 2024-04-11
[WV-337] Convert Form Layout Changes to Include Templates (offices held) TEAM REVIEW

lredmo1:

### DIFF
--- a/office_held/views_admin.py
+++ b/office_held/views_admin.py
@@ -286,9 +286,9 @@ def office_held_edit_view(request, office_held_id=0, office_held_we_vote_id=""):
         }
     else:
         template_values = {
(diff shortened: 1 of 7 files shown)


office_held/views_admin.py
```diff
@@ -232,7 +232,16 @@ def office_held_edit_view(request, office_held_id=0, office_held_we_vote_id=""):
         return redirect_to_sign_in_page(request, authority_required)
 
     messages_on_stage = get_messages(request)
+    district_id = request.GET.get('district_id', '')
```
lredmo1, line 235: Is the `''` correct?

DaleMcGrew: Yes, that's correct! 👍

PR #2446, opened 2024-04-11
[WV-335] Convert Form Layout Changes to Include Templates (office contests) TEAM REVIEW

lredmo1:


office/views_admin.py
```diff
@@ -971,10 +971,21 @@ def office_edit_view(request, office_id=0, contest_office_we_vote_id=""):
     if not voter_has_authority(request, authority_required):
         return redirect_to_sign_in_page(request, authority_required)
 
+    ballotpedia_office_id = request.GET.get('ballotpedia_office_id', 0)
+    ballotpedia_office_name = request.GET.get('ballotpedia_office_name', 0)
```
DaleMcGrew: ballotpedia_office_name should have the alternate value be '' instead of 0.
The same for all of the other incoming values you are adding.

DaleMcGrew (changes requested):

lredmo1:
@DaleMcGrew this has been updated per our conversation and is ready for review
- added dicts to template values so labels still show in the form for `new`
- removed template variables no longer needed now that dicts are used for them

office/views_admin.py
```diff
@@ -956,10 +967,87 @@ def office_new_view(request):
 
     messages_on_stage = get_messages(request)
     template_values = {
+        'ballotpedia_office_id_dict':              
+        {
+            'label':    'Ballotpedia Office Id (Office Held)',
+            'id':       'ballotpedia_office_id_id',
+            'name':     'ballotpedia_office_id',
+            'value':     ballotpedia_office_id
```
lredmo1, line 975: please check that I did these values correctly for all of these. my understanding is that since there is a form for a new office contest, we don't need something like `ballotpedia_office_id if ballotpedia_office_id else office_on_stage.ballotpedia_office_id`

DaleMcGrew:
Thank you @lredmo1! 👍

PR #2450, opened 2024-04-13
[WV-334] Convert Form Layout Changes to Include Templates (candidates) TEAM REVIEW

lredmo1:
Updated so dict values are passed into the form for new offices held (before, labels weren't showing on that form)
Removed the template values that were no longer needed now that dicts were used for them
Updated a few misnamed items

Still need to troubleshoot this error at `http://localhost:8000/c/new/`:

image

### DIFF
--- a/candidate/views_admin.py
+++ b/candidate/views_admin.py
@@ -1731,6 +1731,21 @@ def candidate_new_view(request):
     maplight_id = request.GET.get('maplight_id', "")
     page = request.GET.get('page', 0)
     politician_we_vote_id = request.GET.get('politician_we_vote_id', "")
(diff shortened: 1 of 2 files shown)


lredmo1:
@DaleMcGrew It looks like this error precedes my edits. I get the error at `http://localhost:8000/c/new/` whether on `wv-334` branch or the `develop` branch.

DaleMcGrew:
Hi @lredmo1, the problem is that on candidate/views_admin.py line 1718, we need to wrap the google_civic_election_id and state_code variables with "str", like this:
url_variables = "?google_civic_election_id=" + str(google_civic_election_id) + "&state_code=" + str(state_code)
I can do this -- I just wanted to explain what is causing the crash.

PR #2451, opened 2024-04-13
[WV-337] Convert Form Layout Changes to Include Templates (office held) TEAM REVIEW

lredmo1:
Updated so dict values are passed into the form for `new` offices held (before, labels weren't showing on that form)
Removed the template values that were no longer needed now that dicts were used for them


DaleMcGrew:
👍

PR #2454, opened 2024-04-13
Switch over from PDF2HTML docker installed project to Apache Tika (also installed by docker)

SailingSteve:
So that the Chrome Extension can process PDF forms since PDF2HTML was abandoned
Tika is accurate and fast, but does not retain complex page formatting or links to images like PDF2HTML did.  Tika should be good enough.

Requires Apache Tika to be installed on the production servers
see https://wevoteusa.atlassian.net/jira/software/projects/WV/issues/WV-358


DaleMcGrew:
@SailingSteve I'm sorry for the delay in approving this pull request. If you can rebase these changes so there aren't collisions, I can approve it.

PR #2463, opened 2024-04-15
Fix to fast_retrieve, get total row count from Master server

SailingSteve:
Not from the local developer's PC/Mac!


DaleMcGrew:
👍

PR #2467, opened 2024-04-16
Fix to fast_retrieve, refactored retrieve_controllers into local and master script files

SailingSteve:
Fix to update_fast_load to return status


DaleMcGrew:
👍

PR #2476, opened 2024-04-17
WV-260 Convert methods with warning to @staticmethod and reformat

tejasaditya:


DaleMcGrew:
👍  Thank you @tejasaditya!

PR #2480, opened 2024-04-18
WV-342, add an option to hard delete Voter Guide Possibility entries

stilwellmg:
Added a new option to the "Mark selected row as" drop-down box and the code to process it.


DaleMcGrew:
👍 Thank you @stilwellmg!

PR #2481, opened 2024-04-18
Fix to fast_retrieve, changed to voter_api_device_id

SailingSteve:
Prior use of voter_device_id occasionally worked by accident


DaleMcGrew:
👍

PR #2483, opened 2024-04-21
Fix to fast_retrieve is done

SailingSteve:
candidate_candidatecampaign had grown to so many columns, that we were either crashing or timing out for a 502 on the retrieveSQLTables api request.
Logging to AWS at CloudWatch /ecs/wevote-api does not work for retrieveSQLTables, which made this harder to figure out.


DaleMcGrew:
👍  Thank you for so much work on this @SailingSteve!

SailingSteve:
Resolves 
https://wevoteusa.atlassian.net/issues/WV-207

SailingSteve:
Should also resolve https://wevoteusa.atlassian.net/issues/WV-363

PR #2493, opened 2024-04-29
Another css href url fix #2

SailingSteve:
Also some more info logged for fast load
Cleaned up indenting in api_doc_page.html to reveal previously unclosed div


DaleMcGrew:
👍

PR #2494, opened 2024-04-29
Another css href url fix #3

### LINKED ISSUE #3: Add icon assets
We need to have a starter sprite of the icons that will be used for the app. They don't have to be the ones that will be used when we go into production but we will need some to start with.

SailingSteve:
Try switching over to the (much simplified) Django V5 way of serving static files.


DaleMcGrew:
👍

PR #2497, opened 2024-04-29
Another css href url fix #5

### LINKED ISSUE #5: People Photos
Will need a collection of photos for the cards that will be available for each voter being represented in the proof of concept. This will eventually come from the services but for now we will need static ones. When they are collected, put them into `WebAppPublic/assets/photos/`. This directory has not been created yet.

SailingSteve:
Another path experiment for Django V5 way of serving static files.
If this doesn't work, I'm going to move the static files to the root (like template html files)


DaleMcGrew:
👍

PR #2505, opened 2024-05-02
Another css href url fix #6

### LINKED ISSUE #6: Repo Python/Django Cleanup
The cleanup and organization of this cookie cutter can be cleaned up to make it enhanced for the application. All of the client side code will live in the `WebAppPublic` directory - or whatever this will be renamed (most of the time called public). 

Dale mentioned someone who could be of help to handle this.

SailingSteve:
Django 5, was having problems with static/apis_v1 so changing it to static/v1 
Experiment to move static to root for svgs and css (previously worked locally, but new way for production)


DaleMcGrew:
👍

PR #2507, opened 2024-05-03
Another css href url fix #7

### LINKED ISSUE #7: Create backlog of stories
Will need to have a backlog of stories that we know will be needed for the app to be built. Put these under the Milestone Backlog so we can then pull them into Sprint # Milestone when the story is ready to be worked on

SailingSteve:
I'm thinking it might be the ngnix proxy server setup or new browser strictness, so trying a temporary cdn


DaleMcGrew:
👍

PR #2515, opened 2024-05-06
Css working in production, served from the WebApp

SailingSteve:
Removed temporary cdn links in favor of statically deployed css/js/imgs from WebApp 
These css/js/imgs have only been tested locally, and have not yet been fully tested due to the current deploy problem in production. 
Migrated to Google Maps advanced markers since "As of February 21st, 2024 (v3.56), google.maps.Marker is deprecated" 
Improvement to fast load estimated number of rows function 
Updated the install docs to show using Python 3.11, and to use fast load instead of individual tables (resolves https://wevoteusa.atlassian.net/issues/WV-365)


DaleMcGrew:
👍 Thank you @SailingSteve for your patience.

PR #2516, opened 2024-05-06
Brandta/wv 212

brandta-1:
updated back-end controller to properly update and verify new wikipedia images for candidates, based on their wikipedia page titles


DaleMcGrew:
👍  The political data team will be excited to have this new feature. Thank you @brandta-1!

PR #2520, opened 2024-05-08
WV-291 Convert methods with warning to @staticmethod and fix formatting

tejasaditya:


DaleMcGrew:
👍 Great work @tejasaditya!

PR #2535, opened 2024-05-14
Ready to go: Removed dead copies of css, js, and imgs that have been moved to WebApp

SailingSteve:
To avoid attempts to maintain them in the wrong place. 
Minor fix to Map Points that threw (non-critical) red js exception. 
We have TwitterUser data that does not return a single user, so rather that returning nothing, return the oldest matching twitter user.


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #2539, opened 2024-05-16
WV-349 Updated is_candidate_we_vote_id and is_politician_we_vote_id to return True/False

Almee98:
### WV-349: Update is_candidate_we_vote_id and is_politician_we_vote_id to return True/False - Fixed

Files changed: wevote_functions/functions.py


DaleMcGrew:
Thank you @Almee98! 👍

PR #2550, opened 2024-05-20
More logging to diagnose twitter sign in #3

SailingSteve:
Narrowing in on why the callback url received by the WebApp does not include any query parameters, including the all important oauthToken and oauthVerifier


DaleMcGrew:
👍

PR #2555, opened 2024-05-24
WV-345, add POST method to voter_guide_possibility_position_save.

stilwellmg:
Utilize the new function return_value_from_request so that voter_guide_possibility_position_save_view can process either a GET or a POST method.


DaleMcGrew:
👍 Thank you @stilwellmg!

PR #2556, opened 2024-05-24
Twitter sign-in for WebApp is working (tested with Safari)

SailingSteve:
3 leg OAuth is working again
Tested with Safari locally, should work with all browsers in production
Removed test logging


DaleMcGrew:
👍 Thank you @SailingSteve

PR #2620, opened 2024-07-15
MERGE READY WV-388 candidatesSyncOut does not include any candidates when state_code is na

tejasaditya:
candidatesSyncOut works as expected. It filter by state code of candidate linked to offices in the election, instead of using the state code of office
Also candidate search by election and state is modified to show all national candidates when using state code NA or na


DaleMcGrew:
Thank you @tejasaditya! 👍

PR #2637, opened 2024-07-19
WV-262 converted methods into static methods [MERGE READY]

omgitsmiles:
converted methods from class CTCLApiCounterManager into static methods.


DaleMcGrew:
Thank you @omgitsmiles! 👍

PR #2660, opened 2024-07-30
WV-450 redirected to readonly DB [MERGE READY]

omgitsmiles:
**search_politicians method**

-  Added .using('readonly').all() on a read_only 

> As our conversation on engineering-e slack, search_politicians is a read only method but has no read_only argument passed in the method. Would I need to pass that read_only == true argument in this method?

**retrieve_politician_list**

- Switched read_only argument to True
- Commented out conditional case to be discussed in Team Review

> Conditional statement already written in method, seems to only to be a read only method? 
Would we need conditional case if read_only = True?

### DIFF
--- a/politician/models.py
+++ b/politician/models.py
@@ -1951,6 +1951,7 @@ class PoliticianManager(models.Manager):
 #             politician_entry.delete()
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you @omgitsmiles! 👍

PR #2663, opened 2024-07-31
[WV-430] prevent polititian self-recommendation

smcvey7:
WV-430 is being transitioned to archive for later

In politician/controllers_recommendation.py line 247 I left in a commented-out fix which prevents self-recommendations from being included in the cand_list to be bulk created. This prevents the issue of self-recommendations but leaves many politicians with only 4 related politicians.


DaleMcGrew:
Thank you @smcvey7! 👍

PR #2667, opened 2024-08-02
WV-440 - Fix for Organization model to read-only database

anup2307:
\admin_tools\views.py - Goes to Template values
\issue\controllers.py - data retrieved for Api

\import_export_twitter\controllers.py -  Data is passed on to the below functions
(refresh_twitter_organization_details  -> save_fresh_twitter_details  >save_fresh_twitter_details_to_organization(in this function another instance of Organization is created and the twitter details are updated to the organization table) 

\organization\views_admin.py  - Goes to template values
\organization\models.py - Goes to template values
\position\controllers.py - Organization information retrieved for the voter list template
\voter\views_admin.py - Goes to Template values

\import_export_wikipedia\controllers.py - Added Comment - Cannot be Read-only
\politician\controllers.py - Line - Added Comment - Cannot be Read-only
\import_export_facebook\views_admin.py - Added Comment - Cannot be Read-only

### DIFF
--- a/admin_tools/views.py
+++ b/admin_tools/views.py
@@ -276,7 +276,7 @@ def data_cleanup_organization_analysis_view(request):
             for item in organization_filters:
                 final_organization_filters |= item
 
-            organization_list_with_duplicate_twitter = Organization.objects.all()
+            organization_list_with_duplicate_twitter = Organization.objects.using('readonly').all()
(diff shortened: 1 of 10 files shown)


DaleMcGrew:
Thank you @anup2307! 👍

PR #2680, opened 2024-08-13
WV-451 update uses of ContestOffice to use readonly [TEAM REVIEW]

bonniehuynh:
Updated any uses of ContestOffice.objects.all that could use readonly instead of hitting main database. Also, added error message for google vertex credentials.


bonniehuynh:
made a new PR for WV-451

PR #2685, opened 2024-08-18
WV-451 fixed uses of ContestOffice that could use readonly [TEAM REVIEW]

bonniehuynh:
made changes to ContestOffice to use readonly or added comments where readonly can't be used; also fixed vertex error message to use status instead of print


DaleMcGrew:
Thank you @bonniehuynh! 👍

PR #2693, opened 2024-08-23
[WV-265]Convert Methods with Warnings to @staticmethod in import_export_maplight/models.py

itcreativeusa:
1. Converted methods with "may be static" warnings to @staticmethod in the import_export_maplight/models.py file.
2. Formatted method declarations with wrapped lines to place each parameter on its own line for improved readability.


DaleMcGrew:
👍 Thank you @itcreativeusa!

PR #2701, opened 2024-08-26
[WV-461] offices held loading notice

itcreativeusa:
Added a 'loading' notice at the top of the Offices Held List Page in templates/office_held/office_held_list.html to indicate when the last page has been submitted.


DaleMcGrew:
Looks good. 👍

PR #2702, opened 2024-08-26
[WV-463] measures list page loading

itcreativeusa:
Added a 'loading' notice at the top of the Measures List Page in templates/measure/measure_list.html to indicate when the list page has been submitted.


DaleMcGrew:
👍

PR #2704, opened 2024-08-28
Removed duplicate CharField definition in import_export_maplight/mode…

itcreativeusa:
In the import_export_maplight/models.py file, I removed a duplicate 'models.CharField' declaration that was defined twice with the same parameters.


DaleMcGrew:
👍 Good catch @itcreativeusa!

PR #2713, opened 2024-09-02
Fix NLTK dependency version

josephevans:


DaleMcGrew:
Thank you @josephevans! 👍

PR #2716, opened 2024-09-03
WV-462 [TEAM REVIEW]

ilonagog:
Update representative_list.html with latest changes, create id="representative_search_submit" for Search for Representative and add "loading" notice at top of page when list page has been submitted.


DaleMcGrew:
Thank you @ilonagog! 👍

PR #2725, opened 2024-09-10
[WV-443]Refactor analytics controllers and add batch processing logic.

itcreativeusa:
Added the "move_analytics_info_to_another_voter" function to transfer "AnalyticsAction" records from one voter ID to another.
I tested the function to make sure it moves the correct number of actions and updates the records. I ran tests in the Django shell to confirm that the function works as expected.


DaleMcGrew:
Great job @itcreativeusa! Thank you. 👍

PR #2731, opened 2024-09-11
WV-454 redirected readonly queries of Emailaddress to readonly database [Team Review]

anup2307:
/admin_tools/views.py  - 1064 - goes to one_voter
                                         1347 - goes to one_linked_voter which goes to template


DaleMcGrew:
👍

PR #2732, opened 2024-09-14
WV-389 Add state column and filter by state to endorser view [TEAM REVIEW]

bonniehuynh:
- State is shown in the header for the selected state
- Displays "no candidates" if there are no candidates in the selected state


DaleMcGrew:
👍

PR #2735, opened 2024-09-17
Rework fast load to be adaptable to new table structures

MvPodell:
Old process: convert API JSON to CSV, clean CSV with table-specific changes by row, and then copy into local postgres
- Broke when new columns were added to tables
- Cleaning was hardcoded for each table

New process: convert API JSON to pandas DataFrame, clean without table-specific solutions, convert to CSV, copy into local postgres.
- When new columns are added, no code changes are necessary for table to be cleaned and copied to local postgres
- Cleaning is generalized


DaleMcGrew:
Thank you @MvPodell! 👍

PR #2744, opened 2024-09-23
WV-362 Add New Voter Guide (Admin Tools) has web form glitches [TEAM REVIEW]

Succorro:
New flow for add new endorser:
1. Find Endorser only offers first two form fields
2. If no results found, offer ability to proceed to Create Endorser form
3. Create Endorser
4. Proceed to Edit Endorser page for further edits

Updates to html file logic: 
new logic for form view from search to create via voter_guide/views_admin.py

Updates to create new endorser error handling: 
moved organization_name check up to org results check, previous check would skip over name check and result in errors
removed 'False' from request.POST.get as later input handlers check for not False only


DaleMcGrew:
Thank you @Succorro, if you have any additional changes, we can discuss in a new pull request.

PR #2759, opened 2024-10-04
WV-267 Convert methods with warning to @staticmethod

tejasaditya:


DaleMcGrew:
👍 I appreciate these @tejasaditya!

PR #2789, opened 2024-11-06
[WV-766] Errors encountered after syncing data

itcreativeusa:
1) Added '%Y%m%d'instead of DATE_FORMAT_YMD views_admin.py
2) Added argument self to def process_exception
candidate/views_admin.py
3)Updated office_held_summary URL pattern to accept numeric IDs, added[0-9]
office_held/urls.py

### DIFF
--- a/candidate/views_admin.py
+++ b/candidate/views_admin.py
@@ -1316,7 +1316,7 @@ def candidate_list_view(request):
             # datetime_now = timezone.localize(datetime.now())
             timezone, datetime_now = generate_localized_datetime_from_obj()
             if positive_value_exists(election.election_day_text):
-                date_of_election = timezone.localize(datetime.strptime(election.election_day_text, DATE_FORMAT_YMD)) # "%Y-%m-%d"
+                date_of_election = timezone.localize(datetime.strptime(election.election_day_text, '%Y%m%d')) # "%Y-%m-%d"
                 if date_of_election > datetime_now:
                     time_until_election = date_of_election - datetime_now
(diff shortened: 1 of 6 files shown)


candidate/views_admin.py
```diff
@@ -1316,7 +1316,7 @@ def candidate_list_view(request):
             # datetime_now = timezone.localize(datetime.now())
             timezone, datetime_now = generate_localized_datetime_from_obj()
             if positive_value_exists(election.election_day_text):
-                date_of_election = timezone.localize(datetime.strptime(election.election_day_text, DATE_FORMAT_YMD)) # "%Y-%m-%d"
+                date_of_election = timezone.localize(datetime.strptime(election.election_day_text, '%Y%m%d')) # "%Y-%m-%d"
```
DaleMcGrew, line 1319: Hello @itcreativeusa we want to maintain formatting standards, which is why we are using the constants like DATE_FORMAT_YMD.

wevote_social/middleware.py
```diff
@@ -69,7 +69,7 @@ def __call__(self, request):
 
         return response
 
-    def process_exception(request, exception):
+    def process_exception(self, request, exception):
```
DaleMcGrew, line 72: Instead of adding "self" (which isn't used, you can add `@staticmethod` like this:
Screenshot 2024-11-23 at 9 32 08 AM

DaleMcGrew (commented):
Hi @itcreativeusa thank you for looking for solutions to the problems you saw. I am closing this pull request because we are working on fixing the FastLoad to avoid the template problems you encountered. I added some comments we can discuss in the next meeting.

PR #2790, opened 2024-11-07
[WV-726]Enhanced candidate search to support multi-word queries across multip…

itcreativeusa:
…le fields

- Split  term into individual words to improve search
- Applied individual filters for each word across multiple fields
- Add descriptive comment about search


DaleMcGrew:
Thank you @itcreativeusa I am approving these changes, and I moved WV-726 back to Sprint Ready.

PR #2805, opened 2024-11-20
Fast load was mistakenly overwriting we_vote_id in a few files.

SailingSteve:
Now there is an exclusion list called "do_not_touch" for columns that have UNIQUE constraints, but also have critical data on the master servers.

Fixes https://wevoteusa.atlassian.net/issues/WV-782


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #2816, opened 2024-12-05
WV-465 Made sure loading banner appears when using the search bar and clear search links. TEAM REVIEW

pbnjcub:
Notes: I also added loading banners when "show all states" link appears on some of the pages, but not all. I wanted to wait to confirm with Dale that I should be doing this. Continued work to make form elements across all pages consistent will be completed in issue WV-905

### DIFF
--- a/templates/position/position_list.html
+++ b/templates/position/position_list.html
@@ -89,8 +89,10 @@
         {% endfor %}
         </select>
         {% if state_code != "" %}
-        <a href="{% url 'position:position_list' %}?google_civic_election_id={{ google_civic_election_id }}&candidate_search={{ candidate_search }}{% if show_all_elections %}&show_all_elections=1{% endif %}">
-            show all states</a>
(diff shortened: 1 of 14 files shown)


templates/position/position_list.html
```diff
@@ -210,37 +215,74 @@ <h1>Opinions/Positions/Endorsements</h1>
...
         $('#state_code_id').change(function() {
+            displayLoadingBanner();
             this.form.submit();
         });
+
+        $('form[name="organization_position_choose_election"]').submit(function(event) {
+            displayLoadingBanner();
+        });
+    });
+    $(function() {
```
DaleMcGrew, line 270: Can these function stubs be removed? I understand if you need to leave them in for a future change.

DaleMcGrew (approved):
This looks solid @pbnjcub 👍 I'm merging it now so we can start testing.

PR #2817, opened 2024-12-06
Fastload revamp experiment

SailingSteve:
Moving tables between postgress servers should be easy.  
It didn't work when the two postgres instances had very different versions.  
Now that they are the same, I'm trying a more conventional (and hopefully less problematic approach).  This is the master only side of the solution, with the master dumping the table to a file and saving it to s3.  And then only sending the s3 link to the client.  Once this is deployed in production, I can experiment with the local side.


DaleMcGrew:
👍

PR #2818, opened 2024-12-06
Fastload revamp experiment -- time format fix

SailingSteve:
Microseconds were not working on the production server in the strftime call. But we don't really need microseconds.
No rush, need pg_dump updated first.


DaleMcGrew:
👍

PR #2821, opened 2024-12-10
WV-861 add next page UI [TEAM REVIEW]

itcreativeusa:
1)Implemented backend logic to support pagination with total_count, items_per_page, start_index, and end_index.
2) Added previous_page_url and next_page_url for navigation.
3) Conditionally hid pagination UI when items fit on a single page.
4)Updated the frontend to display "Previous" and "Next" links with the current page number.

### DIFF
--- a/politician/views_admin.py
+++ b/politician/views_admin.py
@@ -564,6 +564,9 @@ def politicians_import_from_master_server_view(request):
 
 @login_required
 def politician_list_view(request):
+    # Pagination parameters
+    page = int(request.GET.get('page', 0))  # Default to page 0
+    items_per_page = 25  # Number of items per page
     # admin, analytics_admin, partner_organization, political_data_manager, political_data_viewer, verified_volunteer
     authority_required = {'partner_organization', 'political_data_viewer', 'verified_volunteer'}
     if not voter_has_authority(request, authority_required):
@@ -1168,7 +1171,34 @@ def politician_list_view(request):
             show_related_candidates=show_related_candidates,
             was_candidate_recently=was_candidate_recently,
             )
+        
+    # Pagination logic
+    total_count = politician_query.count()
+    items_per_page = 25
+    page = int(request.GET.get("page", 1)) - 1  # Page is 0-indexed
+    start_index = page * items_per_page
+    end_index = start_index + items_per_page
+
+    # Fetch only the items for the current page
+    politician_list = politician_query.order_by("politician_name")[start_index:end_index]
+
+    # Determine if there are previous/next pages
+    has_previous_page = page > 0
+    has_next_page = end_index < total_count
+
+    # Update URLs for previous and next pages
+    previous_page_url = f"?page={page}&state_code={state_code}&politician_search={politician_search}" if has_previous_page else None
+    next_page_url = f"?page={page + 2}&state_code={state_code}&politician_search={politician_search}" if has_next_page else None
+    
     template_values = {
+        "start_index": start_index,
+        'politician_list': politician_list,
+        'state_code': state_code,
+        'politician_search': politician_search,
+        'current_page_number': page,
+        'previous_page_url': previous_page_url,
+        'next_page_url': next_page_url,
+        'hide_pagination': total_count <= items_per_page,
         'checkbox_url_variables':       checkbox_url_variables,
         'election_list':                election_list,
         'exclude_politician_analysis_done': exclude_politician_analysis_done,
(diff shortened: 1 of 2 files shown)


templates/politician/politician_list.html
```diff
@@ -336,5 +367,6 @@ <h1>Politicians</h1>
     });
 
 </script>
+<script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
```
DaleMcGrew: Hi Katerenya, we shouldn't have to include this here because jquery is already included in WeVoteServer/templates/template_base.html on line 24. You can upgrade that version in that file if you need a newer version.

politician/views_admin.py
```diff
@@ -1166,7 +1169,34 @@ def politician_list_view(request):
...
+    has_previous_page = page > 0
+    has_next_page = end_index < total_count
+
+    # Update URLs for previous and next pages
+    previous_page_url = f"?page={page}&state_code={state_code}&politician_search={politician_search}" if has_previous_page else None
+    next_page_url = f"?page={page + 2}&state_code={state_code}&politician_search={politician_search}" if has_next_page else None
+    
     template_values = {
+        "start_index": start_index,
+        'politician_list': politician_list,
```
DaleMcGrew: Please add these template variables in alphabetical order, because that helps us make sure we don't duplicate existing variables. (I see "politician_list" and "politician_search" for example are already passed in further down in this template_values dict.)

politician/views_admin.py
```diff
@@ -1166,7 +1169,34 @@ def politician_list_view(request):
...
+        
+    # Pagination logic
+    total_count = politician_query.count()
+    items_per_page = 25
+    page = int(request.GET.get("page", 1)) - 1  # Page is 0-indexed
+    start_index = page * items_per_page
+    end_index = start_index + items_per_page
+
+    # Fetch only the items for the current page
+    politician_list = politician_query.order_by("politician_name")[start_index:end_index]
```
DaleMcGrew: We have already retrieved politician_list above, and augmented it with data from the Representative table. This replaces that work with a brand new query. I think this pagination logic should be included up around lines 992-996 where the current retrieve is happening:

        if not positive_value_exists(show_all):
            politician_list = politician_query.order_by('politician_name')[:25]
        else:
            # We still want to limit to 200
            politician_list = politician_query.order_by('politician_name')[:200]

templates/politician/politician_list.html
```diff
@@ -51,22 +51,24 @@ <h1>Politicians</h1>
 
 <form name="state_code_form" method="get" action="{% url 'politician:politician_list' %}">
     {% csrf_token %}
+    <input type="hidden" name="politician_search" value="{{ politician_search|default_if_none:'' }}" />
```
DaleMcGrew: politician_search is already being included in the form down below on your line 74 in an input, and doesn't need to be added here as a hidden value.

templates/politician/politician_list.html
```diff
@@ -51,22 +51,24 @@ <h1>Politicians</h1>
...
+    <input type="hidden" name="show_all" value="{{ show_all|default_if_none:0 }}" />
```
DaleMcGrew: show_all is already included below as a checkbox, and doesn't need to be included here as a hidden value.

templates/politician/politician_list.html
```diff
@@ -51,22 +51,24 @@ <h1>Politicians</h1>
...
+    <input type="hidden" name="exclude_politician_analysis_done" value="{{ exclude_politician_analysis_done|default_if_none:0 }}" />
```
DaleMcGrew: exclude_politician_analysis_done is already included below as a checkbox, and doesn't need to be included here as a hidden value.

DaleMcGrew (changes requested):
Hi Katerenya, overall you are on the right track. I have these comments for your next iteration. Thank you!

itcreativeusa:
https://github.com/user-attachments/assets/682c1faa-1ef2-4587-9173-1432f1c8c577

itcreativeusa:
This last screen recording shows my last commit : [fix reloading to page 1 if state changed or checkbox]. 
 With these changes, It doesn't go to page 1 or reload if the state was changed or the checkbox checked.

https://github.com/user-attachments/assets/0f62cb7e-03ba-489e-8cab-f73a52b616a2

DaleMcGrew:
👍 Thank you for working through all of the challenges of this page!

PR #2822, opened 2024-12-10
Fastload revamp workable checkin

SailingSteve:
Need to get pg_dump updated on the production server to fully test, but works well locally.


DaleMcGrew:
👍  Thank you @SailingSteve.

PR #2826, opened 2024-12-11
WV-716 - addendum - Ballot Measures - Allow the ability to delete

tejasaditya:
1. Add check for attached positions before deleting measure
2. Show election list at all times when adding/edition a position, even when empty


DaleMcGrew:
👍

PR #2835, opened 2024-12-18
WV-773 Candidate Page - Upgrade delete to require checking box

tejasaditya:
1. Removed single step candidate delete and added checkbox verification
2. Modified return URLs to maintain as much of the original filters as possible


DaleMcGrew:
Thank you @tejasaditya! 👍

PR #2836, opened 2024-12-18
WV-716 Ballot Measures - Allow the ability to delete - Improvement

tejasaditya:
Modified the POST data to use values from current URL rather than measure data. This allows for preservation of original search filters.


DaleMcGrew:
👍  Looks good.

PR #2837, opened 2024-12-18
Changes to get FastLoad working on my machine. Updated to prevent data damage if the process fails for one table.

DaleMcGrew:
Changes to get FastLoad working on my machine. Updated to prevent data damage if the process fails for one table.


DaleMcGrew:
Hi @SailingSteve I'm approving this so you can grab the code more easily. I'm sending you an email with more details. Great work on this new Fastload!!

SailingSteve:
@DaleMcGrew 

The install instructions call out naming the db 'WeVoteServerDB' and they do not have the developer setup an 'admin' account -- maybe instead of changing the template, you should just change your specific environment_variables.json? 

Also using  postive_value_exists to decide if you are going to append to the connection_string will not end up with a valid connection.  So unless that connection string throws a meaningful error calling out the missing variable, it might be better to just check for values and log an error if they are not set.

PR #2838, opened 2024-12-20
Fastload: fixes to displayed timing issues and create_dev_user fix

SailingSteve:
Fixed python manage.py create_dev_user so that it works from the command line again (the problem was due to a major breaking change in a library)

Did not change the environment_variables_template.json, but I can't see how a dev could create a lower case 
WeVoteServerDB, if they were following the instructions.  

I'll rerun from scratch at some point in the future, I think the instructions can be further simplified.


DaleMcGrew:
👍  Thank you @SailingSteve

PR #2846, opened 2025-02-10
WV-707 Add New Candidate Process doesn't have a way to Add if there are potential matches

tejasaditya:
The create new candidate button is always shown and disable/greyed out if requirements are not met which are also shown right alongside the button as long it is disabled. The option is also disabled if when Twitter handle and state code are inputted and the candidate already exists.


DaleMcGrew:
👍  Thank you Tejas!

PR #2863, opened 2025-04-08
[WV-861]Pagination: show previous and next page buttons conditionally…

itcreativeusa:
1) Added 'has_previous_page' and 'has_next_page' variables in politician_list_view.
2) Passed these values in the context to the template.
3) Updated politician_list.html to conditionally display the "Previous" and "Next" buttons.


DaleMcGrew:
👍

PR #2883, opened 2025-06-20
Standardize comments in environment variables template

josephevans:
Use '_comment' key for all comments and avoid using keys that cause exceptions when trying to set environment variables.


DaleMcGrew:
👍

PR #2886, opened 2025-06-24
Increase the maximum size of the data uploads to 20MB

SailingSteve:
The DATA_UPLOAD_MAX_MEMORY_SIZE value in base.py does not seem to make any difference, since the lib/python3.11/site-packages/django/http/request.py file is looking specifically in the settings.py file for this parameter.


DaleMcGrew:
👍

PR #2892, opened 2025-07-05
WV-861: Fix pagination — hide 'Previous page' on first page in footer

itcreativeusa:
Updated the footer pagination in politician_list.html to use has_previous_page
instead of previous_page_url. Now, the Previous Page is only displayed correctly at the Top and Bottom.


DaleMcGrew:
Thank you @itcreativeusa! 👍

PR #2913, opened 2025-09-26
WV-2045: Claim Profile, WeVoteServer: Send other ways to verify [MERGE READY]

Arjunsivakumar28:
In voter_update_view (WeVoteServer/apis_v1/views/views_voter.py) please accept these fields:

politician_we_vote_id

other_ways_to_verify (an assembled Text block with all of the values in the “Verify that you are authorized”)

At the end, the voterUpdate API should return its usual response, plus:

other_ways_to_verify_sent = True

Please note that we aren’t actually updating the voter record, but using this API to email this information


DaleMcGrew:
Great first WeVote pull request @Arjunsivakumar28! 👍

PR #2927, opened 2025-10-15
[WV-2090] Candidate: Data Cleaning, Speedup, and Speed Statistics changes [TEAM REVIEW]

ric0de:
Updated views_data_cleaning.py and its html file to make sure the search filter still works properly without interfering with the background maintenance operations that run on the page. I removed unnecessary search operations that aren't necessary for the data cleaning scripts to appear, which should greatly reduce load times. These changes keep the data cleaning page stable while still allowing the user to search as they would on the main page.

Refactored the candidate_list_view in views_admin.py to make it run faster and cleaner. Combined multiple filter and exclude queries into one using Q() objects to reduce how many times the database is queried (though Django ORM is lazy, still cuts down on repeated evals). I also simplified the search and pagination logic so it’s easier to follow and maintain. Overall, this (along with the views_data_cleaning.py changes) should make the candidate list load quicker without changing how anything looks or works.

Added a template tag (performance_total) to calculate and display the total execution time across all performance snapshots in the speed_statistics_banner.html template. This should work on all instances where speed_statistics_banner is called.

image
image

Apologies! I lost track of time and didn’t consider making separate branches. These changes are related, but next time I’ll plan ahead and split work into separate branches/PRs to keep things cleaner and easier to review.


DaleMcGrew:
Thank you @ric0de! 👍

PR #2930, opened 2025-10-16
[WV-2158] Add "Twitter (click to sort)" link on Candidates page [TEAM REVIEW]

ric0de:
***WARNING***: I'm confident this may have a conflict with my PR for WV-2090. It's only one line, so it should be an easy fix. If that PR makes it through, I'll have to adjust this one.

Notes: Defaults to sorting by candidate name when candidates have the same amount of Twitter followers. Also, candidates with zero Twitter followers are always placed above candidates without Twitters.

Sort by Candidate:
image
Sort by Twitter:
image


DaleMcGrew:
👍

PR #2934, opened 2025-10-20
[WV-2171] Added atomic transaction to 'Link Candidates' for loop

mjacquot1:
Reference: [transaction.atomic() documentation](https://docs.djangoproject.com/en/5.2/topics/db/transactions/)

By wrapping `candidate.save()` and position_list_manager.update_politician_we_vote_id_in_all_positions() inside a `transaction.atomic()` block, each iteration of the loop will queue the database updates into a single transaction. This transaction will only be committed once the `for` loop completes, reducing the overhead compared to committing after every individual operation.

Adding this change has not caused any increase in test failures.

This approach is considered safe under the following assumptions:
1. The live back-end database is PostgreSQL, as it is in our local environment.
2. Both `Politician.save()` and `PoliticianListManager.update_politician_we_vote_id_in_all_positions() `have a low risk of failure. This is reasonable since their core save functionality is unchanged (only field values might be changed), and we haven't been adding explicit error handling.
3. If either function fails within the transaction, we would want all changes to be rolled back and an exception to be raised, preserving atomicity.

### DIFF
--- a/politician/views_admin.py
+++ b/politician/views_admin.py
@@ -12,6 +12,7 @@ from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.contrib.messages import get_messages
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import transaction
 from django.db.models import F, Q
 from django.db.models.functions import Length
 from django.http import HttpResponse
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you @mjacquot1! 👍

PR #2939, opened 2025-10-23
[WV-2180] Added enum values to pass through Politician Edit url redirect

mjacquot1:
Added `politician/enums.py/SpeedStatistics` enum so that we have shorter keys to pass through 
```
HttpResponseRedirect(reverse('politician:politician_edit', args=(
        politician_id,)) + url_variables + "&" + performance_process_dict_encoded)
```

This method of using redirect to pass through speed statistic names and descriptions was erroring since too many arguments in the `GET` request pushed it over the 4,094 character limit. Previously it was sending a `GET` request of ~5,600 characters, and this method of using an enum shrinks it down to ~3,600 characters.

Added `round()` to timestamps to decrease the character count a further ~400 characters to ~3,200 characters total.

### DIFF
--- a/politician/enums.py
+++ b/politician/enums.py
@@ -1,0 +1,163 @@
+from enum import Enum
+
+class SpeedStatistics(Enum):
+    RET_POLITICIAN_FF_ERROR = {
+        "name": "Retrieve politician form fields",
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #2945, opened 2025-10-28
[WV-2135] Split up totals for different funcs in speed_statistics_banner

ric0de:
Previously used to include one grand total even if there were multiple functions. Now it does totals for each individual function


DaleMcGrew:
👍

PR #2946, opened 2025-10-30
[WV-2188] Generate presigned URL on master and pass pull from URl on Local

mjacquot1:
Feature Doc Here: https://docs.google.com/document/d/147RajQvdTI33vLeuT3-SXEoIdV0Vbav_7ynwcEwrVf0/edit?usp=sharing

Previously, the Fast Load process required local AWS credentials to download a database backup file stored in an S3 bucket on the master server. To eliminate this dependency, the master server now generates a presigned URL using generate_presigned_url, which allows secure access to the backup file without local credentials. The local environment has been updated to use Python’s requests library to download the file directly, removing the need to run within an AWS-authenticated instance.

Test Plan:
A test plan need to be devised that best protects data. This PR will be modified to address testing.


DaleMcGrew:
👍 We can discuss in today's meeting. Thank you @mjacquot1

PR #2947, opened 2025-10-30
WV-2105 handle missing/unverified email objects and potential voters linked to unverified emails [TEAM REVIEW]

Arjunsivakumar28:
This is an update for the WV-2105 ticket accept email when creating campaign owner. This covers the following additional cases:

- When an email object with input email address does not exist
- When an email object exists but is not verified
  - (Solution): Create new verified email address object &rarr; create new voter object  &rarr; follow workflow

> [!IMPORTANT]
> Review potential edge case where unverified email exists, and voter object exists connected to unverified email. In this case cannot create new voter object because voter with duplicate emails not allowed.

### DIFF
--- a/campaign/views_admin.py
+++ b/campaign/views_admin.py
@@ -244,25 +244,68 @@ def campaign_edit_owners_process_view(request):
         if re.fullmatch(email_regex, incoming_campaignx_owner_we_vote_id):
 
             # set email address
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍 Thank you @Arjunsivakumar28!

PR #2951, opened 2025-11-04
[WV-2188][quick-fix] create fastload table presigned url

mjacquot1:
Changed the presign URL to have expiration in seconds using `datetime.total_seconds()`. Using a Datetime object was causing an error.
Tested by using Local AWS credentials to create a presigned_url using an existing file:
image
Checked URL for proper expiration:
image


DaleMcGrew:
👍

PR #2954, opened 2025-11-06
[WV-2188] Added Fast Loading with presigned urls and db password

mjacquot1:
Added a `connected_to_aws()` function to verify whether a valid connection can be established with AWS using local credentials. The assumption is that if the local credentials are valid, the code is being executed on the Master server, since most volunteers do not have AWS credentials.

Moved the `tf = tempfile.NamedTemporaryFile(mode='r+b')` call outside of the try block to ensure the temporary file is always properly closed and deleted via `tf.close()`.

Refactored the `pg_restore` command into a list of commands passed to `subprocess.run()`. This approach helps ensure valid commands are passed and allows a database password to be included when required, while avoiding errors when no password is needed. A check for this logic is located at [LN 175].

If an error occurs (typically due to a missing password), it is handled in the exception block.

### Success Measures

Tested by running the Fast Load process and confirming successful synchronization of the local database.

Successful Fast Load after removing AWS credentials from Local
image

Candidate linkages working now for Ted Lieu:
image

Speeds Statistics behaving similar to Live environment:
image

### DIFF
--- a/config/environment_variables-template.json
+++ b/config/environment_variables-template.json
@@ -7,6 +7,7 @@
   "CAMPAIGNS_ROOT_URL":             "http://localhost:3000",
   "CHALLENGES_ROOT_URL":            "http://localhost:3000",
   "TIME_ZONE":                      "US/Pacific",
+  "SERVER_IS_SOURCE_OF_TRUTH":      false,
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
Thank you @mjacquot1 👍

PR #2959, opened 2025-11-14
[WV-2223] Added an Is Claimed Profile filter to Candidate, Politicians and Representatives admin views.

Almee98:
Added an "Is Claimed Profile" filter to Candidates, Politicians, and Representatives admin views.


DaleMcGrew:
Great work @Almee98! 👍

PR #2970, opened 2025-11-26
Fix candidate search 502 timeout and exclude sub-snapshots from Speed Statistics totals

ric0de:
Eliminated the 502 error caused by large candidate searches (~100k+ results) and reverted back from Q-object filtering back to direct query .filter() usage. Even the worst-case queries now complete in roughly 3 seconds on a fast machine.
Also improved nested snapshot display and excluded them from total runtime calcs

Before (buggy, 16s load time): 
image
After (fixed, 3s load time):
image


DaleMcGrew:
I appreciate you following up on this @ric0de! 👍

PR #2971, opened 2025-11-26
[WV-2309] LOCAL Address fast load empty tables

mjacquot1:
This issue was cause by a quirk of a python `tempfile`, where if not enough data is written to it, the buffer wont be hit and the file will technically be empty. On larger files, once the buffer is hit it starts writing to the file. To address this, running `tf.flush()` forces the buffer to write to the file before it is passed into `pg restore`.


DaleMcGrew:
Good catch @mjacquot1! 👍

PR #2976, opened 2025-12-01
[WV-2253] Add scopes, secure user id from leaking on token creation

mjacquot1:
Part of several branches being worked on in concurrence for Fast Load.

Due to no longer needing a direct user connection, some security has been added:

1. Now a scope is required to create and authenticate a token. IE, if we had `Process A` and `Process B`, a token generated for `Process A` cannot be use for `Process B`.
2. Removed returning a user id when a token is created. Now whatever process that is creating the token needs to track the user.

Testing:
Added and modified unit tests to ensure functionality.

### DIFF
--- a/wevote_tokens/models/single_use_tokens.py
+++ b/wevote_tokens/models/single_use_tokens.py
@@ -7,6 +7,10 @@ import json
 from wevote_functions.functions import positive_value_exists
 
 
+class Scope(models.IntegerChoices):
+    NONE = 0, 'none'
+    BACKUP_ONE_TABLE_TO_S3 = 1, 'backup_one_table_to_s3'
+
 class SingleUseToken(models.Model):
     """
     ###################################################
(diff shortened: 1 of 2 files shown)


wevote_tokens/models/single_use_tokens.py
```diff
@@ -66,8 +77,8 @@ class Meta:
     def __str__(self):
         return f"Single Use Token for {self._user_id} (expires: {self._expiration_datetime})"
 
-    # Only modify on creation.
-    def save(self, user_id, validation_key, expiration_seconds=None, json_data=None, *args, **kwargs):
+    ## Only modify on creation.
```
DaleMcGrew, line 80: Hi @mjacquot1 a double ## is not allowed under PEP8. The allowed alternate could be:
# ##

DaleMcGrew (approved):
👍

PR #2980, opened 2025-12-03
[WV-2311] MAIN fast load token validation / issuance

mjacquot1:
@DaleMcGrew DO NOT MERGE YET
Considering what we discussed, I should update `SingleUseToken` separately to change `_user_id` in a way that won' cause migration issues.

Added validation for `SingleUseToken` and issuance in `backup_one_table_to_s3_view`.

Receive token and validation information from requests headers, and return Auth failure or new token info. Right now, this won't really do anything regarding stopping a Fast Load process. However, it will validate and issue new tokens and return the information to the sender.

Testing:
Unit Testing set up for token usage.

PS: Don't know how a `null=True` got snuck into `SingleUseToken._validation` but we certainly don't want it there.

### DIFF
--- a/apis_v1/tests/test_views_backup_one_table_to_s3.py
+++ b/apis_v1/tests/test_views_backup_one_table_to_s3.py
@@ -2,13 +2,14 @@
 from django.test import TestCase
 from unittest.mock import patch, MagicMock
 from django.test import RequestFactory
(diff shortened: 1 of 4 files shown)


mjacquot1:
Refactoring this to play nicer with LOCAL

PR #2985, opened 2025-12-06
[WV-2346] add character field, modify unit tests

mjacquot1:
Adding a string `_user_id` to the `SingleUseToken` model. This should be fine from a security standpoint, as Id values in the column is essentially how Django ORm `ForeignKey` field works.

Modified unit tests to ensure this works.


DaleMcGrew:
👍

PR #2988, opened 2025-12-12
WV-130: Server side to fix which updates supporters_count field in CampignX table on Voter delete

Arjunsivakumar28:
- The supporters_count field in the CampignX table is what the frontend uses to display the support count for politicians on load. This field was not updating on voter delete, so I fixed it under the delete_follow_entries_for_voter() function
- There was also other minor bugs where the support and don't support counts did not go back to zero after reload, so I fixed it at update_or_create_campaignx() function


DaleMcGrew:
Thank you @Arjunsivakumar28! 👍

PR #2990, opened 2025-12-16
[WV-2311] main fast load token authorization issuance

mjacquot1:
Added a `TokensManager` class that can automatically handle the passing of token info and token issuance between functions and views. 

The unit tests can be cleaned up to make it more compact, but honestly I believe having their functionality in is more important than having it look pristine right now.

While the decorator should work as it is thoroughly unit tested, I'm leaving it comment out right now as it would serve no purpose to be live right now.

Expect quick follow up PRs for `Local` -> `Main` token authorization once this is merged into develop.

testing:
Unit Testing


DaleMcGrew:
Thank you @mjacquot1 👍

PR #2992, opened 2025-12-19
Recipient Template List folder structure built

Arjunsivakumar28:
- Added Recipient Template List link in Email Campaigns page (admin side)
- Built Recipient Template list page with structure similar to email template list page
- Create recipient template and edit recipient template links lead to same page


DaleMcGrew:
Thank you @Arjunsivakumar28! 👍

PR #2994, opened 2025-12-26
[WV-2253] local main handshake for fastload

mjacquot1:


DaleMcGrew:
Thank you @mjacquot1 for tackling these security concerns! 👍

PR #2995, opened 2026-01-06
[WV-2253] Comment out login check for fastlaod

mjacquot1:
Currently the token creation method isn't working on MAIN. As for right now, it is working fine on Local, so there is a difference in PROD that is causing this issue. Can't tell what's causing it as it's all try/except statements on MAIN, but I can see from the `response.content` that the log in is in fact working correctly, so this is an issue with tokens.

Likely, I need to send errors to the python logger and view them on cloudwatch in prod.


mjacquot1:
Ah, I believe I found the issue. Nginx drops headers with a period `.` in it.
https://trac.nginx.org/nginx/ticket/629

DaleMcGrew:
👍

PR #2997, opened 2026-01-08
[WV-2253] change header response to json

mjacquot1:
NOTE: I also turned on the decorator for `backup_one_table_to_s3_view` in the interest of expediting functionality testing. This should not have any issue in prod due to all the try/except statements.

The previous issue is that NGINX would remove headers that contained a period `.`
This caused the code to run fine in local, but error in production.

After looking into RFC standards (https://datatracker.ietf.org/doc/html/rfc7230) the best course of action is to send JSON bodies over token headers.

This changes `TokensManager` functions to send and handle headers as JSON now.

Testing:
Tested functionality on local instance
Modified unit tests to handle `JSON` and expect string `datetimes`.


DaleMcGrew:
👍 Thank you @mjacquot1!

PR #3000, opened 2026-01-12
Added voterReviewedApp API to record that a voter has reviewed an app

SailingSteve:
If a negative review was bypassed by the dialogs, send the text of that negative review to Zendesk.


DaleMcGrew:
Hi @SailingSteve I'm approving this and will push it live, but my preference would be to weave these updates into the existing voterUpdate API. 👍

PR #3001, opened 2026-01-14
[WV-2253] require login credentials to begin fastload

mjacquot1:
NOTE: Will not actually block someone from accessing FastLoad on Main. That will be added after monitoring.

Adds back in the initial login credentials requirements for FastLoad. Tested it out and token distribution/authentication is working as expected.

Testing:
video attached
https://github.com/user-attachments/assets/aa64737d-eab5-4c3c-94ca-e87db8110f03


DaleMcGrew:
Thank you! 👍

PR #3002, opened 2026-01-14
Updated the voterReviewedApp API

SailingSteve:
Now receives an email address so that for negative bypass reviews, the voter has the option to receive a written reply via Zendesk.


DaleMcGrew:
👍

PR #3011, opened 2026-01-29
[WV-2461] choose template dropdown email campaigns

mjacquot1:
Purpose: Allow users to select an email campaign template with a directory-like selection tree.

Decided to go with `JsTree` since it was much more suitable to a modular component-like design. I included a comment in the template that tells users what the expected format of the list is.

However, code like this is needed to recieve the tree selection in `selection_tree.html` template in the parent template:
```
{% include "email_outbound/selection_tree.html" with selection_type="some name" selection_tree_data=data %}
document.addEventListener('DOMContentLoaded', function() {
  document.addEventListener('treeSelectionChanged', function(event) {
        someLogic(event)
      });
});
```

Removed HTML and Javascript code that was related to the older way of choosing an email template.
Also fixed an unrelated issue where the email body wasn't loading from the selected template.

Testing:
https://github.com/user-attachments/assets/85338579-3a04-4247-a795-5269edbb7375


DaleMcGrew:
👍 This is really good work @mjacquot1!

PR #3012, opened 2026-01-30
[WV-2484]: Create Test Log View Form (url: /cloudwatch_logging)

Arjunsivakumar28:
- Built a log view form at /cloudwatch_logging where use can chooselog level, action, messages,etc. and send log
- Used to test and check logs in AWS cloudwatch


DaleMcGrew:
👍

PR #3020, opened 2026-02-06
Some defensive changes for voterReviewedApp

SailingSteve:
Might be crashing in production in some circumstances.


DaleMcGrew:
👍

PR #3021, opened 2026-02-07
WV-2447 Add email open-tracking pixel endpoint and recipient fields

Almee98:
- Added tracking pixel endpoint /apis/v1/opened/<open_tracking_code>/ that updates counts/timestamps and returns a 1x1 GIF (204 if missing/unknown code).
- Updated email assembly to generate tracking code before merge and to render a pixel + footer only when a code exists; footer is empty otherwise.
- Added privacy policy link in the footer and used empty alt on the pixel image.
- Added a unit test for the new pixel endpoint in apis_v1/tests/.


DaleMcGrew:
Great work @Almee98! 👍

PR #3025, opened 2026-02-12
add logging to token creation and authentication specifically

mjacquot1:
Adds a logger to creation and authentication specifically. This adds some granularity to logs and will see when someone logs in from Fast Load


DaleMcGrew:
👍

PR #3027, opened 2026-02-13
Logging Nesting on the wrong level.

mjacquot1:
Logging nesting was on the wrong level and broke the if statement.


DaleMcGrew:
👍

PR #3030, opened 2026-02-17
[WV-2500] Email Campaigns: Build Saved Audiences Drawer which can be opened from a few different pages

ric0de:
image

image


DaleMcGrew:
Excellent work @ric0de 👍

PR #3034, opened 2026-02-24
Fix for like/dislike count in candidate page [WV-2304][Merge Ready]

anup2307:


DaleMcGrew:
👍 Thank you @anup2307!

PR #3037, opened 2026-03-02
Fix ballot counts by passing election id list to the backend api

ateek-ujjawal:
WV_839 - Fix ballot measure counts


DaleMcGrew:
Thank you @ateek-ujjawal! 👍

PR #3038, opened 2026-03-03
 [WV-2602] Load Tinymce from webapp files according to production environment structure

Arjunsivakumar28:
- The webapp pull request (https://github.com/wevote/WebApp/pull/4716) puts tinymce files and folders on src/javascript/api-server, which is used by wevote server to load tinymce
- Configs and setup for tinymce done as per production server environment, with a _tinymce_loader_js.html file in templates/email_outbound/js that contains tinymce configs and is included where needed


DaleMcGrew:
Thank you @Arjunsivakumar28!!! 👍

PR #3040, opened 2026-03-04
Search for migrations directories instead of models.py when changing directory ownership.

josephevans:
This will fix our production builds.


DaleMcGrew:
👍

PR #3042, opened 2026-03-04
Checkin migration files

josephevans:
These are the migration files from the live production server.

Going forward, when database schema changes are made, the "makemigrations" command should be run locally and the generated migration file(s) should be included in the PR that changes the schema.


DaleMcGrew:
@josephevans have you had a chance to add the remaining migration files from the former admin server we discussed on Frday August 14th?

josephevans:
@DaleMcGrew Just added the 7 new migration files. As discussed on Friday, I think there are a few files created after we stopped using the admin server for deployments that still need to be generated and added to fully be up-to-date on migration files.

DaleMcGrew:
Thank you @josephevans. I will create a new database on my local machine, grab these migration files, and test running them locally. As you say, I might need to check in more migration files to match what is live on our production database. If everything checks out, I'll approve this pull request, and hopeful the AWS pipeline deployment will just work as-is with these migration files.

josephevans:
@DaleMcGrew I just added a commit to this PR that includes all the latest changes from the `develop` branch. This PR should now include all migration files required.

One thing to note - once this PR is merged, all developers (using docker or not) will have to re-create the database from scratch. This is because the migrations previously created on dev environment setup will not be the same as what is included in this PR.

DaleMcGrew:
I have tested this, and am ready to merge! This will require all developers to upgrade their local WeVote Server and do two things:
Install a brand-new WeVoteServer database
Set up a new cloned WeVoteServer on your local machine (so you can get the updated migration files that have been checked into github, and so they won’t conflict with the migration files you built up locally)

PR #3048, opened 2026-03-13
WV-2502 add attachments to campaigns and email templates and send attachments using sendgrid

Arjunsivakumar28:
- Allows adding attachments to email templates and campaign templates and send emails through sendgrid with attachments
- Attachments added are stored in S3 and referenced when applying templates to mail campaigns
- Currently checks are in place such that max file size is 10mb and specific content types are restricted but these can be changed easily (These globals exist in email_outbound/views_admin.py)
```
MAX_ATTACHMENT_BYTES = 10 * 1024 * 1024  # 10MB; adjust to your needs
ALLOWED_CONTENT_TYPES = {
    "application/pdf",
    "image/png",
    "image/jpeg",
    "text/plain",
    "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
}
```
- EmailAttachments model was created to track attachment files and manage them
- Attachments are read before looping through recipients and passed through functions before sending email via sendgrid


DaleMcGrew:
Incredible work @Arjunsivakumar28. 👍

PR #3049, opened 2026-03-13
Wv 2501: Add inline images to email template and campaign (plus fixes to WV-2502)

Arjunsivakumar28:
- Add inline image attachment feature, which involves fixes in fixes to images_upload_handler function in the tinymce config file (templates/email_outbound/js/_tinymce_loader_js.html)
- preparing inline images differently to send attachment via sendgrid
- More fixes to attachments (like handling template cancels and duplicates appropriately)
- Also 1 caveat, uploading files takes a few seconds every time, which from testing seems to be delays from s3, not the code


DaleMcGrew:
Really great work @Arjunsivakumar28. 👍

PR #3062, opened 2026-03-27
Change SQS process to be a lambda function

josephevans:
Currently, we use a dedicated ECS task to process web functions queued into the SQS queue. This change allows us to create a docker container that can be used as a Lambda function triggered by an SQS queue.


DaleMcGrew:
👍

PR #3063, opened 2026-03-27
Create new lambda function for batch processing

josephevans:
Currently, we have a dedicated ECS task for doing batch processing. This introduces a new lambda handler that can handle batch processing.


DaleMcGrew:
👍

PR #3064, opened 2026-03-28
Update docker install instructions and refactor compose file

josephevans:
To make setting up development environment as simple as possible, making a few changes to the instructions and docker compose file:

1. Use a common `wevote` docker network and add instructions to create it. This allows full-stack devs to run both front and backends locally and have them communicate together.
2. Always run the localstack (AWS simulator) container to simplify the options. We plan to utilize this for the S3 stuff that currently requires hitting a real S3 bucket in the cloud.
3. Update the docker setup documentation to be as simple as possible.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -13,29 +13,21 @@ services:
       retries: 5
     image: postgres:16-alpine
     networks:
-      - back_end
+      - wevote
     ports:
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #3070, opened 2026-04-03
refactor: remove hard-coded AWS credentials, use ECS task IAM role

josephevans:
## Summary

Removes all hard-coded AWS credentials across the codebase, replacing them with ECS task IAM role permissions. Also improves Route53 DNS record handling to prevent accidental overwrites.

## Changes
- `organization/controllers_fastly.py` — use task role for Route53 updates
- `apis_v1/views/views_extension.py` — remove hard-coded credentials
- `email_outbound/functions.py` — remove hard-coded credentials
- `image/models.py` — remove hard-coded credentials
- `retrieve_tables/controllers_master.py` — remove hard-coded credentials

## Additional Fixes
- Check for **any** existing DNS record type (not just CNAMEs) before adding a new Route53 record, preventing accidental overrides of required infrastructure records

## Motivation
Hard-coded credentials are a security liability. Using the ECS task role is the AWS-recommended approach — no secrets to rotate, no risk of leakage, and proper least-privilege access.

### DIFF
--- a/apis_v1/views/views_extension.py
+++ b/apis_v1/views/views_extension.py
@@ -17,11 +17,8 @@ from config.base import get_environment_variable, get_environment_variable_defau
 from exception.models import handle_exception
 from wevote_functions.functions import positive_value_exists
 
-AWS_ACCESS_KEY_ID = get_environment_variable("AWS_ACCESS_KEY_ID")
-AWS_SECRET_ACCESS_KEY = get_environment_variable("AWS_SECRET_ACCESS_KEY")
 AWS_REGION_NAME = get_environment_variable("AWS_REGION_NAME")
 AWS_STORAGE_BUCKET_NAME = "wevote-temporary"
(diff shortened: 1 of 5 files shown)


josephevans:
Updated legacy ECS task permissions so this can be merged now.

DaleMcGrew:
@josephevans can we wait to update this for a couple of weeks while we get the developers up-to-speed on Docker? (If not we lose the ability to use image uploading locally.)

josephevans:
@DaleMcGrew Just FYI - this change should not actually break anything for developers running locally since the boto3 library will look for the credentials in the environment anyways, instead of relying on our code to pass the credentials into the boto3 client.

This change is also required for our prod environment, to avoid having to specify long-lived credentials in the config and instead use IAM role assigned to the running container.

DaleMcGrew:
Thank you @josephevans! Approving it now.

PR #3073, opened 2026-04-08
FastLoad UI improvements so you can tell it is still working

SailingSteve:
All tables seem to load, the we_vote_id in public.candidate_candidatecampaign looks fine.

Confirms fix in https://wevoteusa.atlassian.net/browse/WV-1139


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #3074, opened 2026-04-08
FastLoad UI fixes so progress is more obvious

SailingSteve:
All tables seem to load, the `we_vote_id` in `candidate_candidatecampaign` value looks good.

Further improvements for https://wevoteusa.atlassian.net/browse/WV-1139

Also changed over to Joe's git hash api, instead of scraping.


DaleMcGrew:
👍

PR #3078, opened 2026-04-10
WV-2304: Update like and dislike counts for candidates, politicians and campaign objects so that counts are unified across ballot, candidate search and politician pages

Arjunsivakumar28:
- Added bulk update function refresh_candidate_supporters_count_from_politician in candidate/controllers.py to update candidate counts from politician
- Added refresh_campaignx_supporters_count_from_politician_object in campaign/controllers.py to match campaign and politician counts
- Did not have access to master follow_followorganization table to verify if like/dislike actions update counts correctly, so that needs to be verified


DaleMcGrew:
👍 Thank you @Arjunsivakumar28, I'll add more comments after I play with this on my machine.

PR #3084, opened 2026-04-15
Update docker-based development environment

josephevans:
1. Use a unique namespace for docker compose to avoid having compose think services are orphaned
2. Create `wevote-temporary` bucket in localstack initialization
3. Use log function that flushes output in docker entrypoint script


DaleMcGrew:
👍

PR #3088, opened 2026-04-23
Add pgadmin container to docker compose

mjacquot1:
Runs as expected.

1. `docker compose up --build` to statrt and create the container.
2. Go to `localhost:8080`
3. Default username: `fake_email@wevoteeducation.org`
4. Default Password: `admin`
5. Select `Add New Server`
image
6. Name: `environment_variables.json` value for `DATABASE_NAME`
image

- Host name/address: `db`
- Port: 5432
- Maintenance database: postgres
- Username:  `environment_variables.json` value for `DATABASE_USER`
- Password: Whatever password was used when setup up your postgres superuser as in these instructions: https://github.com/mjacquot1/WeVoteServer/blob/develop/docs/README_API_INSTALL_POSTGRES_MAC.md
- Click `Save`
image

Results:

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -69,6 +69,18 @@ services:
     ports:
       - "127.0.0.1:8000:8000" # this allows access to the WeVote API at localhost:8000
 
+  pgadmin:
+    image: dpage/pgadmin4:8.4
+    environment:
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #3090, opened 2026-04-30
[WV-432] When a candidate or politician entry is saved and choice of photo is changed in the Admin Tools, regenerate background color [Team Review]

zohakhan-wevote:
**PR Description**
**Title**: Fix background color regeneration when photo choice changes in Admin Tools
**Description**:
When a candidate or politician entry is saved and the choice of photo is changed in the Admin Tools, the profile image background color should regenerate to match the new photo. This was broken due to several issues in both _politician/views_admin.py_ and _candidate/views_admin.py._
Root causes fixed:

1. _regenerate_color_ **was always truthy** — It was read from the form as a raw string. An empty string _''_ is not _False_ in Python, so the color was being regenerated on every save even when the admin never clicked the regenerate button. This wiped out the _profile_image_background_color_needed_ flag before it could be saved. Fixed by wrapping with _positive_value_exists_().

2. **Photo type switch path regenerated immediately with wrong photo** — When switching photo type, _generate_background_() was called immediately before the new photo was fully active, generating the wrong color. Fixed by removing the immediate generation and instead setting _profile_image_background_color_needed = True_ to let the edit view handle it on next page load.

3. **Photo delete path never set the flag** — When a photo was deleted, _profile_image_background_color_needed_ was never set to _True_, so the color was never updated. Fixed by adding the flag.

4. **Ballotpedia URL cleared path never set the flag** — Same issue as photo delete. Fixed by adding the flag.

5. _profile_image_background_color_ empty string was overriding the flag — When the color field came in as empty from the form, it unconditionally set needed = False, wiping our flag. Fixed by removing the unnecessary flag assignment from this path entirely.

6. _candidate_edit_view_ was missing the color generation trigger — _politician_edit_view_ already had logic to check the flag on page load and generate the color. _candidate_edit_view_ was missing this entirely. Fixed by adding the same trigger.

**Testing**:
Verified background color updates when switching photo type on a politician
Verified background color updates when switching photo type on a candidate

### DIFF
--- a/candidate/views_admin.py
+++ b/candidate/views_admin.py
@@ -2364,6 +2364,12 @@ def candidate_edit_view(request, candidate_id=0, candidate_we_vote_id=""):
         queryset = queryset.filter(candidate_we_vote_id=candidate_we_vote_id)
         queryset = queryset.order_by('-log_datetime')
         change_log_list = list(queryset)
+        if positive_value_exists(candidate_on_stage.we_vote_hosted_profile_image_url_large):
+            if candidate_on_stage.profile_image_background_color_needed is not False:
+                candidate_on_stage.profile_image_background_color = generate_background(candidate_on_stage)
+                candidate_on_stage.profile_image_background_color_needed = False
+                candidate_on_stage.save()
+                messages.add_message(request, messages.INFO, "Background color generated")
         t1 = time()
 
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #3052, opened 2026-03-17
[WV-775] Updated image handling pipeline [MERGE READY]

nyette:
### Issue(s)

[WV-775](https://wevoteusa.atlassian.net/browse/WV-775)

[WV-775]: https://wevoteusa.atlassian.net/browse/WV-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3091, opened 2026-05-07
[WV-775] Issues with .svg, .tiff, .gif type image files while uploading profile photo [Team Review]

zohakhan-wevote:
Summary
This PR fixes a regression introduced by PR #3052 where animated GIF and WEBP images uploaded as profile photos lost their animation — only the first frame was being saved. GIFs were also being incorrectly converted to WEBP format.

The bug had multiple layers, all in the image processing pipeline:

**Layer 1** — `_voter/controllers.py voter_save_photo_from_file_reader_` (primary cause)

The uploaded image was being processed before reaching the caching code. Two things were happening:

1. GIFs were being converted to WEBP unconditionally:
`# Old broken code
image_types_to_check = ["gif", "tiff"]
if img_dict and img_dict["type"] and any(...):
    image.save(image_data_destination, format="WEBP", save_all=True, loop=0)
    image = Image.open(image_data_destination)
`

2.`ImageOps.exif_transpose()` and `thumbnail()` were being called on all images including animated ones — both strip animation frames. 

By the time the image reached cache_voter_master_uploaded_image it already had only 1 frame.

**Layer 2** — `_image/models.py resize_we_vote_master_image_`
`image.convert()` was being called before checking if the image was animated. convert() strips all animation frames, so even if the master survived with frames intact, the resized versions would lose them.

**Layer 3** — `_image/models.py store_python_image_locally_`
save_all=True was not being passed when saving GIF/WEBP, so only the first frame was written to disk.

**Fix**
`_voter/controllers.py_`

- Removed the unconditional GIF→WEBP conversion block

- Added is_animated check using getattr(image, 'n_frames', 1) > 1

- Skipped ImageOps.exif_transpose() and thumbnail() for animated images since both strip frames

`# New code
is_animated = getattr(image, 'n_frames', 1) > 1

```
if is_animated:
    # Skip exif_transpose and thumbnail — both strip animation frames
    python_image_library_image = image
else:
    python_image_library_image = ImageOps.exif_transpose(image)
    image_copy = python_image_library_image.copy()
    image_copy.thumbnail(
        (PROFILE_IMAGE_ORIGINAL_MAX_WIDTH, PROFILE_IMAGE_ORIGINAL_MAX_HEIGHT),
        Image.Resampling.LANCZOS)`
```
`_image/models.py — resize_we_vote_master_image_`

- Moved media_type check to happen before any convert() call

- Used shutil.copy2 to copy animated files as-is instead of resizing frame by frame (Pillow cannot reliably resize animated images)

- Fixed finally block to add if image: null check

`# New code
if media_type in {'image/gif', 'image/webp'}:
    # Pillow cannot reliably resize animated images frame by frame
    # Copy the original file as-is to preserve all animation frames
    import shutil
    shutil.copy2(original_image_local_path, converted_image_local_path)
else:
    # resize logic for non-animated formats
    ...`
image/models.py — _`store_python_image_locally`_

- Added save_all=True for GIF/WEBP formats
`# New code
if image_format and image_format.lower() in {'gif', 'webp'}:
    python_image_library_image.save(image_local_path, format=image_format, save_all=True)
else:
    python_image_library_image.save(image_local_path, format=image_format)`

**Known Limitations / Follow-up Required**

- Animated GIF and WEBP resized versions are copied as-is from the master (not resized) because Pillow cannot reliably resize animated images frame by frame. The proper long-term fix is to configure the Fastly CDN image optimizer to handle animated formats. A separate ticket should be raised for this.

- CDN should also be reconfigured to leave GIFs as-is and not convert them to WEBP.

### DIFF
--- a/image/controllers.py
+++ b/image/controllers.py
@@ -2296,9 +2296,6 @@ def cache_voter_master_uploaded_image(
     :param voter_we_vote_id:
     :return:
     """
-    # TEMP DEBUG
-    print("DEBUG frames at entry:", getattr(python_image_library_image, 'n_frames', 1))
-    print("DEBUG format at entry:", python_image_library_image.format)
     time0 = log_and_time_cache_action(True, 0, 'cache_voter_master_uploaded_image')
     we_vote_parent_image_id = None
     success = False
@@ -2335,7 +2332,6 @@ def cache_voter_master_uploaded_image(
 
     # image file validation and get source image properties
     analyze_source_images_results = analyze_image_in_memory(python_image_library_image)
-    print("DEBUG frames in analyze:", getattr(python_image_library_image, 'n_frames', 1))
 
     if not analyze_source_images_results['image_url_valid']:
         error_results = {
--- a/image/controllers.py
+++ b/image/controllers.py
@@ -2296,6 +2296,9 @@ def cache_voter_master_uploaded_image(
     :param voter_we_vote_id:
     :return:
     """
+    # TEMP DEBUG
+    print("DEBUG frames at entry:", getattr(python_image_library_image, 'n_frames', 1))
+    print("DEBUG format at entry:", python_image_library_image.format)
     time0 = log_and_time_cache_action(True, 0, 'cache_voter_master_uploaded_image')
     we_vote_parent_image_id = None
     success = False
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
This is a great iteration @zohakhan-wevote, thank you! 👍

PR #3094, opened 2026-05-12
Work towards "Add follow_followorganization to WeVoteServer fastload"

SailingSteve:
It is so long since I worked on this, I wanted the code to be at least queued up for git before my vacation.  Server side works fine with local setup, and in Docker.

### DIFF
--- a/retrieve_tables/retrieve_common.py
+++ b/retrieve_tables/retrieve_common.py
@@ -20,6 +20,7 @@ allowable_tables = [
     'election_ballotpediaelection',
     'election_election',
     'electoral_district_electoraldistrict',
+    'follow_followorganization',
     'issue_issue',
     'issue_organizationlinktoissue',
     'measure_contestmeasure',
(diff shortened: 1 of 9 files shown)


retrieve_tables/retrieve_common.py
```diff
@@ -20,6 +20,7 @@
     'election_ballotpediaelection',
     'election_election',
     'electoral_district_electoraldistrict',
+    # 'follow_followorganization',
```
DaleMcGrew, line 23: Hi @SailingSteve did you mean to uncomment this follow_followorganization line?

SailingSteve: @DaleMcGrew  Yes, since I can't test on the client side until I get it going in SSL -- Needed for "Sign in with Apple" and maybe others.

DaleMcGrew:
👍 Thank you @SailingSteve!

PR #3100, opened 2026-05-14
Changes to reduce politician_retrieve calls and remove row locking (Related to WV-2304 PR)

Arjunsivakumar28:
This PR has fixes for WV-2304. Specifically it reduces politician retrieve calls in refresh_candidate_supporters_count_from_politician() in candidates/controllers.py by calling them in bulk, and removes the for_update in retrieve_politician() function in politician/models


DaleMcGrew:
👍

PR #3101, opened 2026-05-15
Added SSL setup to README_API_INSTALL_DOCKER.md

SailingSteve:
Simplified SSL configuration, it will now start up in the protocol defined in WE_VOTE_SERVER_ROOT_URL  (http or https). 

Created a dummy text file to force the creation of the certs directory. 

There are some seemly random CSRF errors on startup, and while running the app (in both http and https) -- everything seems to work, but I will continue investigating.


DaleMcGrew:
👍

PR #3111, opened 2026-05-28
WV-1001: Fix Follow Topic button not updating followers count on API server

0x416c616e:
## Summary
Fixes two bugs that caused `Issue.issue_followers_count` to not increment when a voter clicks Follow Topic on the Ready page.

1. **Replica lag race** — `voter_issue_follow_view` writes a new `FollowIssue` row to the primary DB, then immediately calls `FollowIssueList.fetch_follow_issue_count_by_issue_we_vote_id`, which queries `.using('readonly')`. If the readonly replica is lagging and doesn't yet have the new row, the stale (pre-write) count gets saved back to the primary, effectively undoing the increment.
2. **Master Sync count/row skew (dev environments)** — Master Sync brings `Issue.issue_followers_count` (e.g. 1,234 followers) but not the underlying `FollowIssue` rows. After clicking Follow locally there is just one row, so the recount-from-rows approach overwrites 1,234 → 1.

## Fix
- `follow/models.py` — `fetch_follow_issue_count_by_issue_we_vote_id` now queries the primary DB (`'default'`) rather than `'readonly'`, since it's invoked immediately after a write.
- `apis_v1/views/views_voter.py` — Instead of recounting all `FollowIssue` rows after the toggle, capture the voter's prior follow state for this issue and apply an atomic `F('issue_followers_count') ± 1` delta. This sidesteps the replica race entirely and is robust against the dev-env count/row skew. The decrement is guarded with `issue_followers_count__gt=0` so the count can't go negative.

## Test plan
- [ ] Signed-out: click Follow on a Ready-page topic → followers count increments on the API server
- [ ] Signed-in: click Follow → increments; click Unfollow → decrements (and does not go below zero)
- [ ] Toggle Follow rapidly → final count reflects final state with no lost updates
- [ ] Dev environment where `Issue.issue_followers_count` ≫ local `FollowIssue` row count → following does not clobber the cached count

Jira: https://wevoteusa.atlassian.net/browse/WV-1001

### DIFF
--- a/apis_v1/views/views_voter.py
+++ b/apis_v1/views/views_voter.py
@@ -9,6 +9,7 @@ from sendgrid import sendgrid
 from sendgrid.helpers.mail import Mail, Email, To, Content
 from django.contrib import messages
 from django.core.exceptions import RequestDataTooBig
+from django.db.models import F
 from django.http import HttpResponse
 from django.views.decorators.csrf import csrf_exempt
 from django_user_agents.utils import get_user_agent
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Great work @0x416c616e! 👍

PR #3112, opened 2026-05-31
Recovered SSL instructions for the README_API_INSTALL_DOCKER.md

SailingSteve:
Changing to SSL used to require lots of changes to environment_variables.json, now you only need to change one new "WE_VOTE_SERVER_PROTOCOL" variable ( set it to 'http' or 'https').

"DEV_SERVER_TYPE" was eliminated. 

Developer's existing variables settings should continue to work as is, given the new default settings, but going forward this is the way to go.

The WE_VOTE_SERVER_ROOT_URL environment variable has been eliminated from the environment_variables.json file, it is now built in base.py from the new variables for the components of the URL: "WE_VOTE_SERVER_PROTOCOL", "WE_VOTE_SERVER_DOMAIN_HTTP","WE_VOTE_SERVER_DOMAIN_HTTPS", and "WE_VOTE_SERVER_PORT".

Some css cleanup on the fast_load_table_statistics() report.

Added documentation about how to use and debug with PyCharm.  Some small config changes were required for this.

Added 'follow_followissue' and 'follow_followorganization' to fast load for https://wevoteusa.atlassian.net/browse/WV-2601
but can't test it until the change too retrieve_common.py gets to the production servers.


DaleMcGrew:
👍

PR #3114, opened 2026-06-05
Added cache creation mechanism and local variables

mjacquot1:
Create cache table in the postgres database. This allows us to use a cache without having to expand our infrastructure.

Following methodology from here: https://docs.djangoproject.com/en/6.0/topics/cache/#database-caching

**Wait to ship.**
Currently only works in DEV, will error on PROD without values added. 

Cons:

    While faster than regular table calls, not nearly as fast as access from a dedicated cache like REDIS.
    Pros:
    Serves as a temporary memory store with a 5 minute expiration of every row
    Can be extended to a dedicated cache machine

Tests:
Unit Tests Created


DaleMcGrew:
👍

PR #3135, opened 2026-06-11
Introduce OpenTelemetry to enable advanced metric collection.

josephevans:
- Add OpenTelemetry SDK and Django instrumentation packages. 
- Add hook to main manage.py script to enable instrumentation.

These changes should be transparent, and simply adds a instrumentation hook for collecting metrics. By default, the metrics are not send anywhere, unless the environment variables `OTEL_SERVICE_NAME` and `OTEL_EXPORTER_OTLP_ENDPOINT` are defined. When these are defined, the metrics will periodically be pushed to the defined endpoint.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -2,7 +2,10 @@ from django.core.management import execute_from_command_line
 import os
 import sys
 
+from opentelemetry.instrumentation.django import DjangoInstrumentor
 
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #3185, opened 2026-06-15
Fix for default WE_VOTE_SERVER_DOMAIN_HTTP and FASTBACK experiments

SailingSteve:
Full test pass for "all" combinations of http and https 
Fastbackup stopped working on the master side, with pg_dump returning empty files.  So added logging, and a new optional EXTENDED_FASTBACKUP_LOGGING true/false variable that can be sent to the master to turn on extended logging -- avoids me having to uncomment and submit PRs each time. 
Commented out the prior fast load approach where data was converted to csv files and then sent by http 
Only for docker/dev/entrypoint, run Python with -u, to allow print lines to end up in the logs


DaleMcGrew:
👍

PR #3187, opened 2026-06-16
Revert back to using NewRelic for application metrics, logging and traces.

josephevans:
After testing out OpenTelemetry logging and metric collection, it seems like NewRelic is better suited for our use-cases and more intuitive to debug problems. In the future, we can revisit this if needed.


DaleMcGrew:
👍

PR #3193, opened 2026-06-23
FastLoad runs in Docker (now including ballot_item)

SailingSteve:
With these changes, all tables now load.


DaleMcGrew:
Thank you @SailingSteve 👍

PR #3203, opened 2026-06-25
[WV-4605] Add cache to politicians performance dict

mjacquot1:
Uses a cache and retrieve the politician performance dict. This removes over 2 thousand characters from the URL.

Testing:
This appears to work with live data now.


DaleMcGrew:
👍

PR #3207, opened 2026-06-26
Update dev dockerfile for security

josephevans:
Update dev docker environment:

1. Remove /tmp mountpoints in containers which was causing permission issues when writing temporary files (pdf2html conversion, for example).
2. Remove `EXPOSE` lines for unused ports from Dockerfile.dev, document ports still left
3. Change bind host for all services to `127.0.0.1`, which is only accessibly locally, instead of 0.0.0.0, which is accessible by anyone on your local network. 
4. Change the port to access pgadmin to 9000, from 8080.

@SailingSteve  @mjacquot1 Please let me know if there is some reason this does not work, and we can look for a solution that does.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -94,7 +94,7 @@ services:
     volumes:
       - pgadmin_data:/var/lib/pgadmin
     ports:
-      - "127.0.0.1:8080:80"
+      - "127.0.0.1:9000:80" # expose pgadmin at http://localhost:9000/
     networks:
       - wevote
 
--- a/compose.yaml
+++ b/compose.yaml
@@ -20,7 +20,6 @@ services:
     user: postgres
     volumes:
       - postgres_data:/var/lib/postgresql/data
-      - /tmp/docker:/tmp                 # Needed to get backup files into psql to test fastback
 
 
   localstack:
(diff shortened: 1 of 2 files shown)


mjacquot1:
As far as I can tell, this shouldn't cause issues. It's currently running fine on my setup, and any references to `0.0.0.0:8000` in the repo seems to be for internal operations in the container.

compose.yaml
```diff
@@ -78,12 +77,11 @@ services:
       - DEV_SERVER_SSL_KEY=${DEV_SERVER_SSL_KEY:-cert/wevotedeveloper.com_key.txt}
     volumes:
       - .:/wevote/code
-      - /tmp/docker:/tmp                 # Needed to get backfiles into psql to test fastback
```
SailingSteve, line 81: I am ok with the port number changes, and the exposure of ports changes.  I did a bunch of experimentation, and probably left the exposure too open.

In controllers_master.py dump_full_postgres_table_to_tmp(), I do a pg_dump to the /tmp directory, these files can exceed 500MB in size.  The pg_dump runs on the Master server in the API container.  In the case where one needs to test fastload on a development machine, where the master and local are the same instance (kind of a mind twister) -- the transfer happens with both ends using the same local postgres -- The table is pg_dump'ed to /tmp, sent to S3,  and in restore_one_file_to_local_server() is brought back down from S3 and stored in the /tmp directory, where it is 'pg_restored into the local db.  In production these two /tmp directories are on different instances, but in testing it is very convenient to have them be in the same directory on the single machine.  I don't believe pg_restore can download directly from S3.

So I am open to suggestions, but I believe that in compose.yaml, that I need in the api container:

volumes:
     - /tmp/docker:/tmp                 # Needed to get backfiles into psql to test fastback

DaleMcGrew:
👍 Thank you everyone for all of the reviews and feedback.

PR #3208, opened 2026-06-27
WV-4654: edited politicians_managed_retrieve_for_api to check for matching emails

Arjunsivakumar28:
- Edited politicians_managed_retrieve_for_api in politician/controllers_managed_politician.py to check for matching emails to identify which politicians voter manages (politiciansManagedRetrieve endpoint)


DaleMcGrew:
👍

PR #3210, opened 2026-07-01
Fix for fastload in to a new or recreated db server, ignore roles from AWS

SailingSteve:
`docker compose down -v`  wipes the entire db instance and all of its data and starts fresh -- that is where this fix comes in.

Also a fix for running FastLoad without Docker for Almee.  I can't test it without Docker, but it works with Docker.


DaleMcGrew:
👍

PR #3213, opened 2026-07-05
WV-4722 Lint & Format Python Code: F401

theflippedbit:
- Ran Ruff rule F401, remove unused imports


admin_tools/admin.py
```diff
@@ -2,6 +2,5 @@
 # Brought to you by We Vote. Be good.
 # -*- coding: UTF-8 -*-
 
-from django.contrib import admin
```
theflippedbit, line 5: This results in an empty (besides comments) `admin.py` file, there's also a few pre-existing empty `tests.py` and `models.py` files that should be cleaned up..

DaleMcGrew:
Thank you @theflippedbit! 👍

PR #3223, opened 2026-07-07
WV-2664: Decouple heart follow/dislike from Choose/Oppose (first-action-wins)

ric0de:
TL;DR: On the backend, your heart (like/dislike) and your Choose/Oppose position no longer quietly overwrite each other. Only your *first* action on a candidate links the two; after that, clicking one leaves the other alone. And if the database hiccups mid-check, the code now plays it safe and preserves whatever you already had instead of risking a wrong flip.

Preserve "first-action-wins" so a heart/dislike or Choose/Oppose click no longer silently flips the voter's other stance:

- organization_dislike / organization_follow only create an Oppose/Support position when none already exists (PositionListManager.positions_exist_for_voter_and_politician).
- voter_opposing_save / voter_supporting_save only auto-dislike/auto-follow the org when the voter has no active heart yet (FollowOrganizationManager.voter_has_active_heart_preference).
- New guard helpers read the primary DB ('default') to avoid replica lag, and fail closed on a DB error so a transient failure can't flip an existing position or heart.

### DIFF
--- a/apis_v1/controllers.py
+++ b/apis_v1/controllers.py
@@ -85,7 +85,14 @@ def organization_dislike(  # organizationDislike
 
     save_oppose_position = False
     if positive_value_exists(make_position_update) and positive_value_exists(politician_we_vote_id):
-        save_oppose_position = True
+        # WV-2664: Honor the docstring intent above ("create an Oppose position IFF a position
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
👍

PR #3229, opened 2026-07-08
Restore center-crop to square for profile image resizing when target dimensions are square

Almee98:
## Summary

Restores square cropping for profile photo resizing when the target output size is square (e.g. 200×200, 48×48, 32×32). Rectangular source images are now center-cropped before resize instead of being fit with `ImageOps.contain`, which preserved aspect ratio and produced non-square output files.

## Problem

Profile photos are displayed as squares in the UI, but non-square uploads (e.g. wide rectangular images) were not being converted to square resized versions. This regressed in PR #3052, which replaced the previous `crop_to_square` logic with `ImageOps.contain` for all non–Facebook-background images.

## Related history

- Regression introduced: PR #3052
- Prior working logic: Jun 2025 jpeg resize fix (`crop_to_square` in `resize_we_vote_master_image`)


DaleMcGrew:
👍

PR #3232, opened 2026-07-13
Wv 4653 implement caching speed stats

mjacquot1:


DaleMcGrew:
Thank you @mjacquot1, we are lucky to have you! 👍

PR #3240, opened 2026-07-17
WV-2671 Add unsubscribe link to politician campaign emails

lyndabanh:
* Add politician_email_subscription_secret_key to Politician and EmailCampaignRecipient models
* Run migrations on both models
* Update augment_email_campaign_recipient() to populate politician secret key as fallback when no EmailAddress-based key exists, generating one if needed
* Inject unsubscribe link containing email subscription secret key into email footer in merge_email_campaign_recipient_with_template()
* Add politiciancampaign modifier to unsubscribe_instant_view()


DaleMcGrew:
Thank you @lyndabanh! 👍

PR #3242, opened 2026-07-22
Pass AWS keys into boto3 for image S3 and square-crop static WebP/GIF.

Almee98:
## Summary
- Pass `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` from `environment_variables.json` into boto3 (via `get_environment_variable` + `positive_value_exists`) for image S3 upload/download/delete, so local/dev no longer depends on process env or `~/.aws` for those calls.
- Square-crop **static** WebP/GIF the same way as other still profile images; only **animated** WebP/GIF skip resize (and avoid `SameFileError` when src/dest paths match).
- Set `resized_image_created = True` only on successful resize/copy (not unconditionally after failures).

## Context
Follow-up to the profile square-crop work on WV-4728. Hitting `store_image_to_aws failed: Unable to locate credentials` because boto3 was created with region only. Static WebP/GIF were also skipped entirely by the animated copy path.

## Tested
- [x] Wide JPEG / PNG → `_200x200` / `_48x48` / `_32x32` are square
- [x] Static wide WebP → square sizes (no SameFileError)
- [x] Static wide GIF → square sizes
- [x] Animated WebP/GIF → remain non-square (expected for now)

### DIFF
--- a/image/models.py
+++ b/image/models.py
@@ -43,6 +43,16 @@ WIKIPEDIA_IMAGE_NAME = "wikipedia_image"
 
 AWS_REGION_NAME = get_environment_variable("AWS_REGION_NAME")
 AWS_STORAGE_BUCKET_NAME = get_environment_variable("AWS_STORAGE_BUCKET_NAME")
+AWS_ACCESS_KEY_ID = get_environment_variable("AWS_ACCESS_KEY_ID", no_exception=True)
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you @Almee98! 👍

PR #3244, opened 2026-07-25
add header options to otel

mjacquot1:
Add options for Prod to begin gathering metrics on headers.

Sensitive headers are sanitized.

NEXT:
- Add X-Ray trace ids
- Implement whitelisting for POST


DaleMcGrew:
👍

PR #3248, opened 2026-07-28
resolved uncaught exception state

mjacquot1:
Moved down if statement to ensure `datetime` is changed to string.


DaleMcGrew:
👍 Thank you @mjacquot1. I should get this on the live site in the next couple of hours.

PR #3258, opened 2026-08-07
Crop and resize animated GIF/WebP frames to square profile sizes.

Almee98:
## Summary
- Crops and resizes each frame of animated GIF/WebP profile uploads so large/medium/tiny outputs are square (same rules as still profiles), instead of copying the original rectangular animation as-is.
- Preserves per-frame duration and loop when re-encoding with Pillow `ImageSequence` + `save_all`.
- Completes the WV-4728 animated follow-up after stills / static WebP-GIF square crop.


DaleMcGrew:
👍 Thank you @Almee98!

PR #3261, opened 2026-08-09
Return 400 and 500 codes when authentication fails for fastload

mjacquot1:
Instead of always allowing for Fastload to run, now it requires proper authentication headers. If authentication fails, it will prevent fastload from working and error on Local.

Testing:
Unit tests added for expected failure responses.


DaleMcGrew:
👍 Thank you @mjacquot1.

PR #3267, opened 2026-08-12
Fix local syncing, and handle empty headers in the tokenmanager

mjacquot1:
Accidentally broke syncing on `LOCAL`, but `PROD` is secure. Fixed that, and I also handle empty headers in `TokensManager` now to prevent unnecessary database calls.

Testing:
Added unit tests.


DaleMcGrew:
👍

PR #3268, opened 2026-08-14
Add password reset for WeVoteServer sign in (WV-4734)

0x416c616e:
This adds a password reset to the WeVoteServer sign in for staff, developers, and volunteers, built on Django's password reset and the forms in django.contrib.auth.forms (PasswordResetForm and SetPasswordForm). The emailed link is single use and expires after an hour, and there is no account creation in this flow.

It is careful about a few things: the request page looks the same whether or not the address has an account (and takes the same amount of time), so it can't be used to find accounts, and the link is built from the configured server URL rather than the request Host so a forged Host can't poison it. Rate limiting is per IP with a cascading ban rather than a hard per email cap, so nobody can lock someone else out of resetting their own password. All the counting uses the existing cache table, so there is no new migration.

It also turns on Django's password validators (AUTH_PASSWORD_VALIDATORS), which were not configured before, so weak passwords can no longer be set anywhere, with a minimum length of twelve.

The reset email goes out through whatever email backend is configured, which is SendGrid in production, so nothing new is needed there. The console backend is only for local testing, where the link prints to the terminal instead of being sent.

Tests are in admin_tools/tests/test_password_reset.py.

### DIFF
--- a/admin_tools/tests/test_password_reset.py
+++ b/admin_tools/tests/test_password_reset.py
@@ -1,0 +1,262 @@
+# admin_tools/tests/test_password_reset.py
+# Brought to you by We Vote. Be good.
+# -*- coding: UTF-8 -*-
+
+"""
+Tests for the WeVoteServer password reset flow (staff, developers and volunteers).
+See admin_tools/views_password_reset.py.
+
+These lean on the behavior we actually care about: an attacker must not be able to learn which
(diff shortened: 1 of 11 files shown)


DaleMcGrew:
Thank you @0x416c616e! 👍

PR #3274, opened 2026-08-21
Make sure only unexpired tokens are checked.

mjacquot1:
This should reduce some of the redundant SQL calls. This makes sure only unexpired tokens are checked for verification.

Testing:
Added unit tests.


DaleMcGrew:
👍

PR #3275, opened 2026-08-21
Updates to make deployment pipelines more stable

josephevans:
- Update buildspecs to not use intermediate image tag
- Stop using latest tag for images
- Add retry logic for updating Lambda code via ECR.


DaleMcGrew:
👍

PR #3279, opened 2026-08-25
WV-4523 Email Campaigns: Upgrade "Search recipients" to support multiple word search

theflippedbit:
* Removed duplicate code
* Moved JS out of HTML into own source
* Moves CSS out of HTML into own source
* Added counts to tab labels
* 'AND' logic for filtering
* Added clear 'x' in input


DaleMcGrew:
Beautiful refactoring and solid fix @theflippedbit, thank you. 👍

PR #3282, opened 2026-08-30
WV-4936 Fix manually-entered JIRA subtasks being tied to file columns

xcarolan:
Manual subtask descriptions (checkbox unchecked, 1-4 typed names) were still being matched positionally to file columns 14-17. If the uploaded file didn't have that many columns, or a row's cell was blank, the manual subtask was silently dropped for every story - surfacing as "enter 4 subtasks, only 3 get created."

Manual mode now creates one subtask per entered description on every story, independent of the file's structure. File-driven mode (names auto-detected from headers) is unchanged: still reads task_url from the matching column and skips a story when that cell is blank.

### DIFF
--- a/import_export_jira/controllers.py
+++ b/import_export_jira/controllers.py
@@ -358,6 +358,7 @@ class JiraExcelLoader:
                 f"Consider splitting into smaller batches to avoid timeout or memory issues."
             )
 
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Thank you @xcarolan 👍

PR #3286, opened 2026-09-12
Make sure the passed email is the email_to

mjacquot1:
Problem: Regardless of the valid email passed into the form it will send the reset email to the primary account email.

Solution: Pull the value from the form and make sure that it is passed into the `send_mail()` function

Testing:
Testing is tricky because of how much of it is using django abstracted auth classes. I can confirm through debugpy that this function is called. when the send reset email is sent. I'm fairly confident that it is pull the email value form the form. But this will need to be verified in Prod.


DaleMcGrew:
Thank you @mjacquot1! 👍

